PR #1, opened 2023-09-25
New template parser with validation

uguraslan:
This is the complete rewrite of template parser. It provides following improvements and changes:

- Added separate parser functions for different tasks
- Now the cursor moves in chunks by using simpler regular expressions
- Empty start & end tags are now supported (`<></>`)
- Text in between an opening and a closing tag is captured and can be accessed as `._innerText`. But the text must not contain `<` char.
- Parser functions now works in a complex sequence as middleware and on each cycle cursor position is checked (that approach prevents many problems might occur when the template definition is complex)
- Attributes in nameless tags are no longer supported and the test case related to that has been commented out.
- Main parser function has own output format and there a formatter function to produce required output format for the code generation.
- The template syntax is checked while parsing operation and a custom `Error` is thrown when there is a syntax/format related problem. The error also shows where the parsing error occurred. 
- While formatting the code it also checks if there is a closing tag for each opening tag at the same level. And it throws an error if there is a mismatch.
- Attribute values can be defined by using both single or double quotes.
- Attribute values can contain any character except the used delimiter (if attribute defined like this `x='example'`, the value cannot contain a single quote char `'` or vice versa)
- Attribute names can contain characters defined in this regex: `[A-Za-z0-9:.\-_@]`
- Tag names can contain only following chars: `[a-zA-Z0-9_\-.]` (nameless tags is a special case)


michielvandergeest (approved):
LGTM

PR #2, opened 2023-09-26
Text component failed to render 0

suresh-gangumalla:
[Issue]
Found Text component failed to render 0 (numeric) value

[Reason]
Logical OR statment in computed method 'text' in Text component

[Fix]
Added check for zero value for both slotContent and content

### DIFF
--- a/src/components/Text.js
+++ b/src/components/Text.js
@@ -60,7 +60,11 @@ export default () =>
     ],
     computed: {
       text() {
-        return this.slotcontent || this.content || ''
+        return (
+          this.slotcontent ||
+          this.content ||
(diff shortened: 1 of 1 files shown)


michielvandergeest:
hey @suresh-gangumalla thanks for this PR. You are right that the conditional in the Text component fails when the content has a falsy value.

While your PR fixes that case, I think it can better be fixed in a different place (at the code generator level). The `feature/inline-text-support`-branch (WIP) approaches it a different, and should properly accept falsy values for text content. So I'll close this PR, in lieu of that upcoming change.

Thanks in any case for bringing this bug to the attention!

PR #8, opened 2023-10-07
Fix renderer function name changes and bump renderer tot 0.3.6

MLangendijk:


michielvandergeest (approved):
LGTM. Thanks for the contribution!

PR #21, opened 2023-11-14
Fixes some template parsing bugs and improves validation

uguraslan:
- Enforces having only one container tag at the top level of templates. It can be a single self-closing tag (without no other tags) or a classing opening/closing tag pairs encloses all the other tags.
- Fixed a bug that prevents parser to correctly identify non-closed opening tags.
- Checks also if there is any closing tags without an opening tag.
- Fixed a bug that causes the parser to incorrectly mark `</Tag/>` or similar tags as valid self-closing tags.
- Added more tests (mainly testing the cases where the parser must fail due to invalid templates)
- `format()` function is replaced with a better one.
- moved `parseLoop()` into the try-catch block and `format()` function was moved before.

### DIFF
--- a/src/lib/templateparser/parser.js
+++ b/src/lib/templateparser/parser.js
@@ -19,7 +19,7 @@ import symbols from '../symbols.js'
 
 class TemplateParseError extends Error {
   constructor(message, name, context) {
-    super(`TemplateParseError ${message}`)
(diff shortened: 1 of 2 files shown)


erikhaandrikman (approved):

michielvandergeest (approved):
LGTM!

PR #22, opened 2023-11-20
Use correct "watch" key for watchers docs

robbertvancaem:


michielvandergeest (approved):
LGTM. Thanks for spotting this one Robbert :)

PR #23, opened 2023-11-21
Feature/custom shaders

jfboeve:
Base.js

- removed shaders object
- added shaderAlias object containing pointers to other shader types.
- use the renderer's shadermanager to check which shader types / effects are available.

Depends on:

https://github.com/lightning-js/renderer/pull/89


jfboeve:
@michielvandergeest this can be merged now

PR #25, opened 2023-12-13
Feature/create app 2

sandeep-vedam:
Added support for creating the Lightning 3 app.


michielvandergeest (changes requested):
Thanks for the PR so far Sandeep. I've left some comments inline.

I also noticed that when I cancel the creation flow in the middle (ctrl+c) it throws an error instead of gracefully handling it.

sandeep-vedam:
> Thanks for the PR so far Sandeep. I've left some comments inline.
> 
> I also noticed that when I cancel the creation flow in the middle (ctrl+c) it throws an error instead of gracefully handling it.

Hey @michielvandergeest , I have handled this case. Please have a look.

michielvandergeest (changes requested):
While testing I encountered the following issues:

- if I select to _not_ initialize a git repository, the app is created but the flow is hanging
- if I select no to enable eslint (but yes to git) I get an error (`handleErrorFromBinding(ctx); Error: ENOENT: no such file or directory, mkdir`)

michielvandergeest:
Linking to issue #43

PR #28, opened 2024-01-07
Fixes unescaped single quotes issue for content values

uguraslan:
Unescaped single quotes in the `content` attribute of a `Text` tag lead to syntax errors in the code generated by the `cast` function:

```html
<Text content="Let's play" />
```

creates a problem whereas 

```html
<Text content="Let\'s play" />
```

works without an issue.

The fix checks if there are any unescaped single quotes in the `content` value and escapes it properly.


michielvandergeest (approved):
LGTM!

PR #32, opened 2024-01-14
Improved Error Handling for Template Parser

uguraslan:
- Template parsing errors are separated from structural errors. For structural issues like unclosed tags or having multiple root-level elements, a new error type called `TemplateStructureError` has been introduced.

- Now, the error messages give more details. They show the component name, the parent component (when precompile is off), and the component's file path (when precompile is on). This helps to know where the error is happening.

- Template parsing and structural errors are critical and can no longer be suppressed. So, encountering these errors will now halt the application to ensure that those issues are addressed.

- When an error occurs, the error message will also show where in the template the problem (roughly) is. This makes it easier to find and fix the issue.

- Fixes an issue where closing tags with attributes were incorrectly accepted. Now, the parser can catch this kind of problem too.

#### When precompile is off
Screenshot 2024-01-14 at 14 50 30

#### When precompile is on 

Screenshot 2024-01-14 at 14 56 14

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -47,11 +47,11 @@ const required = (name) => {
 }
 
 const Component = (name = required('name'), config = required('config')) => {
-  const setupComponent = (lifecycle) => {
+  const setupComponent = (lifecycle, parentComponent) => {
     // code generation
     if (!config.code) {
       Log.debug(`Generating code for ${name} component`)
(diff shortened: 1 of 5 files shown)


michielvandergeest (approved):
LGTM!

PR #34, opened 2024-01-22
Resolved issue with setting mount, pivot and scale values to zero (0) in object literal with x and y key

suresh-gangumalla:
Resolved issue with setting mount, pivot and scale values to zero (0) in object literal with x and y key

### DIFF
--- a/src/element.js
+++ b/src/element.js
@@ -128,11 +128,11 @@ const Props = {
   },
   set mount(v) {
     if (typeof v === 'object' || (isObjectString(v) && (v = parseToObject(v)))) {
-      if (v.x) {
+      if (v.hasOwnProperty('x')) {
         this._props.mountX = v.x
         this._set.add('mountX')
       }
-      if (v.y) {
+      if (v.hasOwnProperty('y')) {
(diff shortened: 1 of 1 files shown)


src/element.js
```diff
@@ -128,11 +128,11 @@ const Props = {
   },
   set mount(v) {
     if (typeof v === 'object' || (isObjectString(v) && (v = parseToObject(v)))) {
-      if (v.hasOwnProperty('x')) {
+      if (Object.prototype.hasOwnProperty.call(v, 'x')) {
```
michielvandergeest, line 131: Maybe we should use `in` instead `hasOwnProperty`. I think it's marginally faster, but it's definitely more readable

michielvandergeest (changes requested):
Good catch Suresh!
I left a comment on a different way to get to the same result

src/element.js
suresh-gangumalla: Agreed @michielvandergeest .  Replaced _hasOwnProperty_ check with _in_

michielvandergeest (approved):
LGTM

PR #36, opened 2024-01-24
Initialized timeout and interval Id's to empty

suresh-gangumalla:
Initializing timeout and interval IDs property on every component instance to empty array ([])

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -148,6 +148,9 @@ const Component = (name = required('name'), config = required('config')) => {
       },
     })
 
+    this[symbols.timeouts] = []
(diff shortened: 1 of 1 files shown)


michielvandergeest (changes requested):
That's not exactly what I meant @suresh-gangumalla. I think we should remove the intervals and timeouts completely from the base object. And instead add those snippets to the `defineProperties` in `component.js`. Similar to how we define the props for each component instance

suresh-gangumalla:
Got it @michielvandergeest I will update PR

michielvandergeest (approved):
LGTM

PR #37, opened 2024-01-25
implemented prepublishOnly script for npm publish (compiling components)

uguraslan:
- Compiles components and generates code for all framework components while running `npm publish`
- Uses `prepublishOnly` life cycle script in `package.json` which runs only during `npm publish`

_We have some eslint errors & warnings regarding generated code and I am not sure if we should address them with this PR or with a separate PR._


michielvandergeest (commented):
Hey Ugur, thanks for setting this up! :)

I have moved the compiler code into the `src` folder. That seems like a more logical place than having this core functionality in a `scripts` folder. The pre-publish script itself is fine inside a dedicated scripts folder, of course

In my tests it the built-in component didn't seem to get compiled. My guess it's because they don't match the `source.indexOf('Blits.Component(') > -1 || source.indexOf('Blits.Application(') > -1` check.

Finally on the eslinting: do we need to care about this? It's supposed to be compiled code, and it's only happening in a `dist` version of the npm package?

uguraslan:
Hey @michielvandergeest, I fixed the double comma problem for the generated effects code. Also checked `context` variable but I saw some references in the code generator and I think it would be better to keep it for now. So I disabled ESLint's unused variable checking for the generated code. Unused variables do not cause errors, ESLint just warns about them anyway.

michielvandergeest:
This looks great Ugur!

One tiny little additional request: it would maybe be nice to _also_ keep a reference to the original js file in there (with the template). Maybe something like `Cirlce.raw.js`

And one question: when it runs the prepublish hook, does it precompile and save the change is the `src` directory? Or is that a temporary location NPM uses to prepare the files to be published?

uguraslan:
> One tiny little additional request: it would maybe be nice to _also_ keep a reference to the original js file in there (with the template). Maybe something like `Cirlce.raw.js`

Sure, I will do that.

> And one question: when it runs the prepublish hook, does it precompile and save the changes in the `src` directory? Or is that a temporary location NPM uses to prepare the files to be published?

According to the documentation: _Runs BEFORE the package is prepared and packed_. So it means that it operates on the actual `src` directory, not in a temporary location.

uguraslan:
I pushed new changes:

- refactored prepublishOnly script so it does not traverse the entire folder structure to match given paths with `componentDirs`.
- prepublishOnly script now creates a backup copy of the original component file. (`Circle.js` => `Circle.orig.js`)
- added a postpublish script and it restores original files from the backup and deletes backup files.

michielvandergeest:
Linking to issue: #42

michielvandergeest (approved):

PR #50, opened 2024-02-06
Filling route path with dynamic params on backKey

suresh-gangumalla:
If dynamic params are exists in a route then replacing dynamic paths with their respective values


michielvandergeest (approved):
LGTM

Issue #54, opened 2024-02-06
Enable option to pass data when navigating to new route


michielvandergeest:

michielvandergeest:
Added in v0.7.0

PR #61, opened 2024-02-19
Added functionality to pass data while navigating, which is injected …

michielvandergeest:
…as a prop - breaking change on the router.to() method.


michielvandergeest:
Addresses issue #54

PR #71, opened 2024-02-22
MSDF font generator

uguraslan:
- generates MSDF font file on-the-fly by using Vite dev server middleware
- if the build command is executed without previously launching the development server and accessing the application in a browser, MSDF fonts corresponding to the TTF fonts in the `public/` directory might not be generated yet. To fix this issue, the plugin now makes sure that MSDF files are generated for all TTF fonts within the public/ directory that do not have corresponding MSDF files. 
- copies generated files to the build output directory after build is completed.


uguraslan:
Converted it to draft, because I am going to improve the build process.

uguraslan:
Now it is ready

PR #72, opened 2024-02-23
multi component support for precompiler

### LINKED ISSUE #65: Pre-compile files with more than 1 componts
The pre-compiler currently only compiles the first Blits component in a file. 

While it can be considered convention / good practice that each component has it's dedicated file, it's a perfectly valid use case to have more than 1 component per file.

We should adapt the pre-compiler plugin to detect this use case and pre-compile all components.

uguraslan:
Implements multi component support (in a single file) for the precompiler library.
Fixes #65

### DIFF
--- a/src/lib/precompiler/precompiler.js
+++ b/src/lib/precompiler/precompiler.js
@@ -25,40 +25,46 @@ export default (source, filePath) => {
     source.indexOf('Blits.Application(') > -1 ||
     /=>\s*Component\(['"][A-Za-z]+['"],/s.test(source) // blits component
   ) {
-    // get the start of the template key in de component configuration object
-    const templateKeyRegex = /template:\s*([`"'])*/g
-    const templateStartResult = templateKeyRegex.exec(source)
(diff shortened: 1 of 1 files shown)


michielvandergeest:
Looks great!

With regards to commented-out components: this is still throwing an error for me when I include this for example:

```
// const bla = Blits.Component('Bla', {
//   components: {
//     PortalRow,
//   },
//   template: `
//     <Element w="1920" h="1080" color="red"></Element>
//   `,
// })
```

uguraslan:
Hey @michielvandergeest, I copied the commented out code you provided at the end of `src/pages/Colors.js` and tested again but I did not get any errors. I also checked the compiled source code on the browser side and the precompiler did not add `code` section to it. Can you please check it again?

Screenshot 2024-02-27 at 11 50 29

michielvandergeest:
Yeah I think it was a local caching issue on my end. This is working as expected.

michielvandergeest (approved):

PR #73, opened 2024-02-26
Improve CoreExtension functionality 

jfboeve:
Added a way to extend renderer functionality without having to use the import chunk url plugin. This means we won't have to rely on vite to build a blits project.

also added a setting called `extensionLoader` which should replace the `fontLoader` setting for launching Blits.


michielvandergeest (approved):
LGTM

PR #77, opened 2024-02-29
Added clearIntervals & clearTimeouts functions

suresh-gangumalla:


michielvandergeest (changes requested):
This looks good! Thanks @suresh-gangumalla.

I do see we now have some duplication in the code. The exact same for loops are in the destroy method. Maybe we should call the $clearIntervals and $clearTimeouts methods in the destroy method?

suresh-gangumalla:
Done @michielvandergeest.   Replaced destroy method  clear timeout and intervals logic with $clearIntervals and $clearTimeouts

michielvandergeest (approved):
LGTM

Issue #76, opened 2024-02-29
Rename ‘function’ key in transition object to ‘easing’


michielvandergeest:
Currently the transition object can have a ‘function’ key to specify a custom timing function. In retrospect this name isn’t as descriptive as I wanted it to be. Also, using the key ‘function’ in the template can throw off the code highlighting in the VS code plugin.

In order to address these 2 issues, we should  rename the key ‘function’  to ‘easing’

Before: 

```js
<Element :y.transition=“{v: $y, function: ‘ease-in-out’}” />
```

After:
 
```js
<Element :y.transition=“{v: $y, easing: ‘ease-in-out’}” />
```

This will be a breaking change. As we are in beta, I don’t think we need to follow a full deprecation path. But we should log a descriptive error when the old ‘function’ key is used in a transition object, to make it straightforward to make the necessary fixes in the app code.

sandeep-vedam:
This issue is addressed with #78 . Hence closing this.

PR #78, opened 2024-03-01
Renamed transition function key to easing key

sandeep-vedam:
This PR fixes the issue #76

### DIFF
--- a/src/element.js
+++ b/src/element.js
@@ -307,6 +307,7 @@ const Element = {
       easing:
         typeof transition === 'object'
           ? 'function' in transition
+          ? console.log("function key is deprecated, please use easing key") : 'easing' in transition
             ? transition.function
             : 'ease'
           : 'ease',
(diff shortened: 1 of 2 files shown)


src/element.js
```diff
@@ -307,6 +307,7 @@ const Element = {
       easing:
         typeof transition === 'object'
           ? 'function' in transition
+          ? console.log("function key is deprecated, please use easing key") : 'easing' in transition
```
michielvandergeest, line 310: Let's use the built in Log component here, instead of a plain console.log. Also we can improve the debug message a bit: "The property for defining a transition's easing function has been renamed from `function` to `easing`. Please update your code."

michielvandergeest (changes requested):
I think there is a missing part in this PR. We have deprecated the `function`, but I don't think it actually got replaced by the new key `easing` ...

michielvandergeest (approved):

PR #79, opened 2024-03-01
Processing color values in the template parser

### LINKED ISSUE #68: Pre-process hardcoded colors in the template (in the parser)
Currently blits normalizes named, hexedecimal and rgba colors to the renderer specific `0x000000ff` format at runtime using a normalize colors function.

For dynamic values we will continue to use that, but for hardcoded colors in the template we can do this ahead of time in the template parser step.

### LINKED ISSUE #70: Add missing testcases to template parser
The testcases for the template parser have some missing coverage. Seems mostly related to Error handling.

Whenever we have time it would be nice to add tests for this as well and get (as close as we can) to 100%

uguraslan:
- modifies `colors.normalize()` method with very strict checks for possible color code/values.
- adds an optional function argument (`defaultColor`) to `colors.normalize()`, which is used by the template parser.
- the template parser now returns template objects with normalised color codes/values. It does that by checking if an attribute value is a color code/value or if the attribute contains any color code/value when the attribute name is one of the following: `color`, `:color`, `:effects`, `effects`. (fixes #68 )
- `color.normalize()` always returns a string unless the optional `defaultColor` is set to something like `null`. If numeric values for colors are needed, it should be done at later stages after color values are normalised as strings. 
- The HTML color values are converted to strings.
- The tests related to `colors` library have been updated.
- The tests for the template parser have been updated & fixed. Now the coverage is 100%. (fixes #70 )

Example:

```html
<Element :color="$backgroundColor">
  <Element color="{top: '#44037a', bottom: '#240244'}" />
  <Element color="#44037a" />
  <Element color="{top: '#44037a'}" :effects="[$shader('radius', {radius: $radius})]" />
  <Element color="transparent" :effects="[$shader('border', {width: 20, color: '#60a5fa'})]" />
  <Element color="#fba" /> 
</Element>
```  

The template parser outputs an object. But to visualise how it process colors, the following can be considered the equivalent of the template above after the template parser processes color values:

```html
<Element :color="$backgroundColor">
  <Element color="{top: '0x44037aff', bottom: '0x240244ff'}" />
  <Element color="0x44037aff" />
  <Element color="{top: '0x44037aff'}" :effects="[$shader('radius', {radius: $radius})]" />
  <Element color="transparent" :effects="[$shader('border', {width: 20, color: '0x60a5faff'})]" />
  <Element color="0xffbbaaff" /> 
</Element>
```


sandeep-vedam:
I have verified the part of PR and here is the coverage for before this fix and after this fix

| S.No | Code               | Before Fix | After Fix |
| ---- | ------------------ | ---------- | --------- |
| 1    | lib/templateparser | 94.16      | 100       |
| 2    | lib/setup          | 39.11      | 40.04     |
| 3    | lib/reactivity     | 34.67      | 34.67     |
| 4    | /lib/colors        | 97.89      | 100       |
| 5    | lib/codegenerator  | 100        | 100       |
| 6    | lib                | 70.09      | 70.09     |
| 7    | components         | 34.22      | 34.22     |
| 8    | source             | 36.66      | 35.46     |
| 9    | announcer          | 55.42      | 55.42     |


| S.No | Desc   | Total Tests | Tests Passed | Tests Failed | Comments                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
| ---- | ------ | ----------- | ------------ | ------------ | ------------------------------------------------------------------------------- |
| 1    | Before | 218         | 188          | 30           |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
| 2    | After  | 239         | 238          | 1            | Generate code for a template with a simple for-loop on an Element with a key attribute |

src/lib/colors/colors.js
```diff
@@ -19,52 +19,62 @@ import htmlColors from './htmlColors.js'
...
+  // The defaultColor parameter should be set to null when utilizing this function within the template parser.
+  // Because the template parser evaluates attribute values, even if they are not color codes,
+  // to determine whether they represent color values or not.
+  normalize(color = '', defaultColor = '0xffffffff') {
     color = color.toString()
 
-    if (color.startsWith('0x')) {
-      return color
-    }
+    if (color.length > 0) {
```
michielvandergeest, line 28: I think this snippet may need to be added back in. All colors, even those that have been normalized during the build step, will still pass through the normalize function at runtime.

So if we do an early return of the ones starting with '0x', then that will save on the execution.

michielvandergeest (changes requested):

src/lib/colors/colors.js
michielvandergeest: Discussed offline and the early return does happen on line 36 (after verifying it's an actual valid color)

michielvandergeest (approved):
👍

PR #83, opened 2024-03-28
Implemented router before hook

suresh-gangumalla:
The router before hook has access to router path parameters in the order they are defined in the router path.

The following cases are covered under router before hook

- If the before hook results in a string, navigate to the respective route.
 `  {
      path: '/xyz',
      component:  Xyz,
      hooks: {
        before: async () => {
          //check for authorization/token verification
          return '/abc';
        },
      },
    },`
- If the before hook results in object, injecting object properties as props in component
` {
      path: '/about/:type/:id',
      component: About,
      hooks: {
        async before({ type, id }) {
          const data = await getDetails(type, id)
          return { data };
        },
      },
    },`

### DIFF
--- a/src/application.d.ts
+++ b/src/application.d.ts
@@ -115,6 +115,11 @@ declare namespace Application {
      * politeness level
      */
     announce?: String | RouteAnnounce
+    /**
+     * Register hooks for the route
+     */
+    hooks?: object
+
   }
 
   export interface ApplicationInstance extends ComponentInstance {}
--- a/src/application.d.ts
+++ b/src/application.d.ts
@@ -93,7 +93,7 @@ declare namespace Application {
   }[keyof T]
 
   interface RouteHooks {
-    before?: (to?: Route, from?: Route) => any;
+    before?: (to: Route, from: Route) => string | Route;
   }
 
   type Route = {
--- a/src/application.d.ts
+++ b/src/application.d.ts
@@ -1,3 +1,4 @@
+
 /*
  * Copyright 2023 Comcast Cable Communications Management, LLC
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -91,6 +92,10 @@ declare namespace Application {
     [K in keyof T]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<keyof T, K>>>
   }[keyof T]
 
+  interface RouteHooks {
+    before?: (to?: Route, from?: Route) => any;
+  }
+
   type Route = {
     /**
      * URI path for the route
@@ -118,8 +123,13 @@ declare namespace Application {
     /**
      * Register hooks for the route
      */
-    hooks?: object
-
+    hooks?: RouteHooks
+    /**
+     * Allows adding route path parameters to the params property internally 
+     */
+    readonly params?: {
+      [key: string]: string | number;
+    };
   }
 
   export interface ApplicationInstance extends ComponentInstance {}
--- a/src/application.d.ts
+++ b/src/application.d.ts
@@ -125,11 +125,21 @@ declare namespace Application {
      */
     hooks?: RouteHooks
     /**
-     * Allows adding route path parameters to the params property internally 
+     * Route path parameters
      */
     readonly params?: {
-      [key: string]: string | number;
-    };
+      [key: string]: string | number
+    }
+    /**
+     * Allows for attaching custom data to a route, either hardcoded in the
+     * route definition or asigned to the route object in a before hook
+     *
+     * Will be merged with the route params and navigation data and passed as
+     * props into the route component
+     */
+    data?: {
+      [key: string]: any
+    }
   }
 
   export interface ApplicationInstance extends ComponentInstance {}
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -118,6 +118,7 @@ export const navigate = async function () {
         if (route.hooks.before) {
           beforeHookOutput = await route.hooks.before(route, previousRoute)
           if (isString(beforeHookOutput)) {
+            currentRoute = previousRoute
             to(beforeHookOutput)
             return
           }
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -111,19 +111,19 @@ export const navigate = async function () {
   if (this.parent[symbols.routes]) {
     const previousRoute = currentRoute
     const hash = getHash()
-    const route = matchHash(hash, this.parent[symbols.routes])
+    let route = matchHash(hash, this.parent[symbols.routes])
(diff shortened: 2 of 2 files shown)


src/router/router.js
```diff
@@ -108,8 +111,20 @@ export const navigate = async function () {
     const previousRoute = currentRoute
     const hash = getHash()
     const route = matchHash(hash, this.parent[symbols.routes])
-
+    let beforeHookOutput
     if (route) {
+      if (route.hooks) {
+        if (route.hooks.before) {
+          const args = route.params ? route.params : {}
+          beforeHookOutput = await route.hooks.before(args)
```
michielvandergeest, line 119: I would not declare a constant for this. Instead we can probably just inline `route.params || {}` inside `before`-function.

In fact, does it actually happen that params is not an empty object? Since on L79 params are reduced from an empty object literal

src/router/router.js
```diff
@@ -108,8 +111,20 @@ export const navigate = async function () {
...
+          if (isString(beforeHookOutput)) {
+            to(beforeHookOutput)
+            return
+          }
+        }
+      }
+      beforeHookOutput = isObject(beforeHookOutput) ? beforeHookOutput : {}
```
michielvandergeest, line 126: what does this line do exactly? I'm not sure I understand :)

src/router/router.js
```diff
@@ -135,7 +150,12 @@ export const navigate = async function () {
         holder.set('w', '100%')
         holder.set('h', '100%')
         // merge props with potential route params to be injected into the component instance
-        const props = { ...this[symbols.props], ...route.params, ...navigationData }
+        const props = {
+          ...this[symbols.props],
+          ...route.params,
+          ...navigationData,
+          ...beforeHookOutput,
```
michielvandergeest, line 157: let's think carefully about the order here .. what should take precedence? `navigationData` or `beforeHookOutput`?

michielvandergeest (changes requested):
Looks good, thanks for contributing this! 

I've left a few questions / remarks in the PR

src/router/router.js
```diff
@@ -108,8 +111,20 @@ export const navigate = async function () {
...
     const route = matchHash(hash, this.parent[symbols.routes])
-
+    let beforeHookOutput
     if (route) {
+      if (route.hooks) {
+        if (route.hooks.before) {
+          const args = route.params ? route.params : {}
+          beforeHookOutput = await route.hooks.before(args)
+          if (isString(beforeHookOutput)) {
+            to(beforeHookOutput)
```
michielvandergeest, line 121: We should probably pass in the data and options as well in the redirect

src/router/router.js
suresh-gangumalla: Will remove constant declaration.
If a route path does not contain any dynamic route paths` route.path.indexOf(':')== -1`, then the params key is not available for such routes.

src/router/router.js
suresh-gangumalla: If the beforehook returns something other than an object, such as a function or a number, then map the beforeHookOutput to an empty object so that it won't be a problem with spread operator

src/router/router.js
michielvandergeest: Yeah, you're right. Maybe we should fix this in the `matchHash` function instead then. That seems like a cleaner solution.

src/router/router.js
suresh-gangumalla: I believe beforeHookOutput should take the highest priority because navigationData is provided at the time of declaring route definition, whereas beforeHookOutput can be the result from an API call.  

But I feel the best could be, users can still safely provide different keys in the beforeHookOutput object and navigationData object, ensuring that everything is available in options

src/router/router.js
suresh-gangumalla: In that scenario, we can support two variants of redirections. 

- The first one is if beforeHookOutput is of type string, `return "/abc"` then ignoring data and option arguments in the router.to() function. 

- The second one is if beforeHookOutput is an object with the 'path' key being mandatory to represent the path to the target/redirect route, along with 'data' and 'options' keys for respective data.  `return {path: "/abc", data: {a: 1, b:2}, optiosn: {inHistory: true}`
      -  But a major problem with the second variant comes in differentiating between API data/object keys that are expected as props in the component. Here, we should strictly follow the rule that the existence of the 'path' key in the object is treated as redirection, otherwise, it is considered as an object for injecting into component props

src/router/router.js
suresh-gangumalla: @michielvandergeest  I have verified all possible types with spread operator like `{...1}`, `{...()=>{}}`, `{...undefined}` and found three is no effect with these values so we can safely ignore this checking. I will remove this statement.

src/router/router.js
michielvandergeest: Yeah, I agree with your logic! :)

michielvandergeest:
As discussed today offline, I’ve been thinking a bit about this feature and perhaps the before hook should receive 2 arguments: the route object for the route we’re navigating **to** and the route we’re navigation **from** (```before(to, from)```)

This allows for doing something conditional in the before hook based on where we’re navigating from.

The second benefit is that we can do changes to the route object we’re navigating to (i.e. inject a different component, add / change props, etc.) and return that object as the result of the before hook. If the route object is returned it will be used in the next step of the router.

For basic redirection to a different route, I guess we can continue to return a string

Please take a look at how feasible this change is and whether there are things unclear or unexpected issue arise.

src/application.d.ts
```diff
@@ -115,6 +115,11 @@ declare namespace Application {
      * politeness level
      */
     announce?: String | RouteAnnounce
+    /**
+     * Register hooks for the route
+     */
+    hooks?: object
+
```
michielvandergeest, line 122: we should probably specify what hooks are available, and the signature of those hooks (currently only before)

src/router/router.js
```diff
@@ -107,9 +111,19 @@ export const navigate = async function () {
...
-    const route = matchHash(hash, this.parent[symbols.routes])
-
+    let route = matchHash(hash, this.parent[symbols.routes])
+    let beforeHookOutput
     if (route) {
+      if (route.hooks) {
+        if (route.hooks.before) {
+          beforeHookOutput = await route.hooks.before(route, previousRoute)
+          if (isString(beforeHookOutput)) {
+            to(beforeHookOutput)
```
michielvandergeest, line 121: The redirect works great! One small thing is that we shouldn't store the route we are redirecting from in history, because then the back functionality is broken.

michielvandergeest (approved):
LGTM!

PR #86, opened 2024-04-12
Updates TypeScript type definitions

uguraslan:
- corrected type definitions for `ComponentInstance.$log` methods
- added `$router` types
- added `$announcer` types
- improved IntelliSense suggestions for `this` in other `ComponentConfig` methods or inside state object. Keys or method names of `props`, `state` will be suggested when `this.` typed.
- props name suggestions are only available when advanced prop format is not used.
- updated `application.d.ts` also to fix some issues.

_The changes are initial improvements for what we already have, not a final and completely correct type definitions_

### DIFF
--- a/src/component.d.ts
+++ b/src/component.d.ts
@@ -81,7 +81,14 @@ declare namespace Component {
   }
 
   interface Input {
-    [key: string]: (this: ComponentInstance, event: KeyboardEvent) => void,
+    [key: string]: ((this: ComponentInstance, event: KeyboardEvent) => void) | undefined,
+
+    /**
+     * Catch all input function
+     *
+     * Will be invoked when there is no dedicated function for a certain key
+    */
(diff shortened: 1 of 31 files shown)


src/component.d.ts
```diff
@@ -74,31 +82,25 @@ declare namespace Component {
 
   interface Input {
     [key: string]: (this: ComponentInstance, event: KeyboardEvent) => void,
-    /**
-     * Catch all input function
-     *
-     * Will be invoked when there is no dedicated function for a certain key
-     */
-    'any'?: (this: ComponentInstance, event: KeyboardEvent) => void,
```
michielvandergeest, line 82: I think this one shouldn't be removed. This is not a TS `any`, but there is an actual key method called `any` that acts as a catch all key handler ..

uguraslan: Oh, you are right. I will get it back.

michielvandergeest (approved):
LGTM! :)

PR #87, opened 2024-04-17
Destroy hook in children components not invoked

suresh-gangumalla:
While navigating away previous router view will get removed, as part of it invoking destroy method on previousView but before that making all children elements/Components  to null leads to failure of setting state of children component to destroy so respecitve destroy hook in that component won't execute


michielvandergeest (approved):
LGTM

PR #89, opened 2024-04-22
Remove event listeners of previous route component

suresh-gangumalla:
Removing event listeners of previous route component before creating target(next) route component instance.


michielvandergeest (changes requested):
As discussed offline:

I think the responsibility of removing listeners does not lie with the router, but should be in the component. As the component already does this clean up, we should probably find a way to indicate to the component to (temporarily) not handle events, via a method or a property on the component.

This removes the need for having to re-register event listeners for when a previous page has the option to keep the page in memory. Also, since pages can stay in memory we need a flag or method to unset the "ignore events" behaviour.

PR #92, opened 2024-05-07
Animation cleanup

michielvandergeest:


suresh-gangumalla:
The animate related code changes are working as expected without any issues. 
I have verified the changes against TMDB and other transition related sample applciation and not landed into any transition issues..

wouterlucas (approved):
Nice! Though the events are about to go in, you could already move this along and revisit once https://github.com/lightning-js/renderer/pull/274 lands or hold out for it.

Issue #94, opened 2024-05-16
window.parent overwritten when using :for directive


WoodingMP:
When writing Blits components, if you use `:for` to generate elements from an array, window.parent will be overwritten with a Blits component, when it should be a reference to itself or its actual parent - see ([`window.parent`](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent)). 

### Reproducability
The issue can be found on any page that uses a `:for` loop. To reproduce the issue:

1. Go to the [blits demo](https://blits-demo.lightningjs.io/#/) and open developer tools
2. Go to the sources tab and add a watcher to `window.parent`
3. Reload the page, and after it's loaded, refresh the watcher
4. Notice that `window.parent` is set to a Blits component

<img width="1531" alt="image" src="https://github.com/lightning-js/blits/assets/44930483/345b7a06-1af0-4287-adef-0a93c7ada17f">

5. Visit any page that doesn't use a `:for` loop (e.g. [intro](https://blits-demo.lightningjs.io/#/demos/intro))
6. Refresh the watcher and notice that `window.parent` is correctly set to a reference of `window`

<img width="1531" alt="image" src="https://github.com/lightning-js/blits/assets/44930483/34594d2b-bfd8-42c2-b131-f1a828064b20">

### Playground
I've also setup a playground to show specifically that it is the :for property that is causing this. Open both playgrounds and view the console, in the ready hook, the window and window.parent will be logged and you'll notice in the playground with the for loop, window.parent is logged as a Blits component.
[...]

PR #95, opened 2024-05-16
Fixed issue with parent being set on the global window object (fixes …

michielvandergeest:
…#94).


michielvandergeest:
Fix confirmed to work by issue opener in #94

PR #102, opened 2024-05-30
Updated SDF and Web Font Construction

suresh-gangumalla:


src/engines/L3/fontLoader.js
```diff
@@ -16,10 +16,24 @@ export default () => {
...
       }
       stage.fontManager.addFontFace(
-        new SdfTrFontFace(font.family, {}, font.type, stage, font.png, font.json)
+        new SdfTrFontFace(font.type, {
+          fontFamily: font.family,
+          descriptors: {},
+          atlasUrl: font.png,
+          atlasDataUrl: font.json,
+          stage,
+          ...(font.metrics ? { metrics: font.metrics } : {}),
```
michielvandergeest, line 25: can't we just do:

```
metrics: font.metrics || {}
````

Comment (suresh-gangumalla) [reply src/engines/L3/fontLoader.js:25]: I tried this @michielvandergeest  Getting problem at this line of code in renderer
https://github.com/lightning-js/renderer/blob/main/src/core/text-rendering/font-face-types/TrFontFace.ts#L150

Comment (suresh-gangumalla) [reply src/engines/L3/fontLoader.js:25]: can I change it to `metrics: font.metrics || null `

**Comment by uguraslan** [comment]
In the `package.json`, I think the latest version of `renderer` should be used to make the changes work.

**Comment by suresh-gangumalla** [comment]
> In the `package.json`, I think the latest version of `renderer` should be used to make the changes work.

yes @uguraslan . I will do that update package.json

File: src/engines/L3/fontLoader.js (hunk 1 above)
Comment (uguraslan) [reply src/engines/L3/fontLoader.js]: @suresh-gangumalla Actually, it can be 

```js
metrics: font.metrics
```

 directly because the renderer code checks it with an if statement (as you pointed out), so if `font.metrics` does not exist (`undefined`) it will processed as the same with `null` in the renderer code.

suresh-gangumalla: Got it @uguraslan 
You are right. we can't just pass an empty {} but instead `undefined ` is fine

uguraslan (approved):
LGTM

michielvandergeest (approved):
LGTM

PR #103, opened 2024-05-30
Updated fonts in boilerplate

suresh-gangumalla:
Removed existing .json and .png file and replace with respective font face .ttf file. Vite plugin will create required .json and .png using lightningjs/msdf-generator for requested .ttf file


michielvandergeest (approved):
LGTM

PR #108, opened 2024-06-10
props tests added

M4tiz:


michielvandergeest (approved):
Great! Thanks for adding these missing test cases

PR #109, opened 2024-06-11
Cleaning up old msdf font files

uguraslan:
#### Old MSDF Font File Detection:

- The plugin now scans all font files in the `public` folder and the temporary folder for automatically generated MSDF files.
- It identifies MSDF font files and checks if the `lightningMetrics` property exists in each file.
- If `lightningMetrics` is missing, it indicates that the file was not generated using the Lightning MSDF generator tool.

#### Temporary File and Variable:

- A temporary file is created to keep track of the cleanup process.
- A variable is introduced to prevent running the same cleanup process multiple times.

#### Warning Visibility:

- The function call for the cleanup process is intentionally placed within the default plugin function.
- This ensures that the warning messages are visible in the terminal output without the need to scroll up.
- Placing the function call outside the default plugin function would cause Vite to clean up the terminal output, hiding the warnings from the view.


vite/msdfGenerator.js
```diff
@@ -3,6 +3,93 @@ import * as fs from 'fs'
...
+            }
+          }
+        }
+      }
+    }
+  })
+}
+
+const cleanup = () => {
+  if (fs.existsSync(path.resolve('.', '.msdf-cleanup'))) {
```
uguraslan, line 56: @michielvandergeest I am not sure about having this. If people miss the warning on the first time, they can't see it again unless the temp file is removed. On the other hand, removing this will cause scanning all paths again on each `npm run dev`.

michielvandergeest:
Closing in favour of a simpler approach ([#1d21c23](https://github.com/lightning-js/blits/commit/1d21c237f6bcaa9620d324d94fd85fab6dbd5f01))

PR #110, opened 2024-06-13
component tests added

M4tiz:


michielvandergeest (changes requested):
@M4tiz thanks for opening this PR. Great work!! ❤️ 

The tests look good to me. 2 remarks:

1) it throws an import error of launch.js, which has been addressed already on the dev branch - maybe we should bring this branch up to date with that one before merging in
2) when I run the full test suite, the tests related to component id are failing. In isolation the tests do pass. Probably related to the global counter already being used in previous tests 

Would you be able to take a look at these 2 points? Thanks again!

M4tiz:
@michielvandergeest Thanks for checking. Rebased on most recent dev branch.

The thing with tests is that we don't have a mechanism of resetting the app state after each test run. I added another method to `componentId.js` that resets the counter. It's not ideal solution, since usually we don't want to add production code only to be used by tests, however it fixes the issue for now. Let me know if this is fine to you :) or maybe you have other suggestions

michielvandergeest:
@M4tiz yes I agree that it's not so nice to add a method just for the tests. I don't suppose it would work if we export the `let counter` and reset it manually in the tests itself (instead of a dedicated reset counter method)?

michielvandergeest (approved):
LGTM

PR #111, opened 2024-06-14
Make focus hold timeout configurable

M4tiz:


src/focus.js
```diff
@@ -48,8 +49,7 @@ export default {
             focusChain = []
           }
         },
-        // todo: make the hold timeout configurable?
-        this.hold ? 50 : 0
+        this.hold ? Settings.get('focusHoldTimeout') : 0
```
michielvandergeest, line 52: Here we need to keep the default of 50 when the setting is not provided by the user. We can do this by passing a second argument to `Settings.get()`

boilerplate/js/default/src/index.js
```diff
@@ -6,6 +6,7 @@ Blits.Launch(App, 'app', {
   h: 1080,
   multithreaded: false,
   debugLevel: 1,
+  focusHoldTimeout: 50,
```
michielvandergeest, line 9: I would prefer to not add this setting to the boilerplate and let the defaults do their work, unless there is a specific need to adjust this timeout

src/launch.d.ts
```diff
@@ -243,7 +243,16 @@ export interface Settings {
...
+  renderMode?: RenderModes,
+
+  /**
+   * The delay, in milliseconds, after the key is released before the selected component
+   * gains focus. This delay can be used to create a smoother transition between focused
+   * components or to prevent rapid focus changes.
+   *
+   * Defaults to `50` (ms)
+   */
+  focusHoldTimeout?: number
```
michielvandergeest, line 255: I'm in doubt about the name and the description. Maybe `holdTimeout` is more fitting. What the value does is configure how long a user should hold down a key, until we consider it to be in a `hold` state. And when in a hold state Blits will prevent focus being passed around and only does until the key is lifted again.

michielvandergeest (changes requested):
@M4tiz thanks for adding this '"todo" functionality 🙂 

I've left some comments in the review. Also I think this new configuration value should be applied in `src/application.js` in the `keyDownhandler`

src/launch.d.ts
M4tiz: I believe you are describing focus handling in `application.js` which is a bit different logic than the one in `focus.js`.
In the screen capture attached I set `focusHoldTimeout` to `1000 ms`. As you can see this timeout configures when the focus is triggered on the selected component _after_ the button is released.

Do you still want to rename it? Also do you want the same config to control the focus handling in `application.js`?

src/focus.js
M4tiz: Done

boilerplate/js/default/src/index.js
M4tiz: Done

src/launch.d.ts
michielvandergeest: that's correct, they are not exactly the same but they _are_ related I think, in the sense that they should probably hold the same value.

@DouweCnossen can you confirm this from a UX perspective?

michielvandergeest:
I have discussed this offline with @DouweCnossen and we came to the conclusion that in all cases we could think of both timeout values should match. So let's for now make this into a single configuration setting (I think `holdTimeout` makes most sense).

If in the future a valid use case comes up where these values may need to be different we can reconsider adding a second (optional) launch setting for that

michielvandergeest (approved):

PR #113, opened 2024-06-18
Routes definition validation

M4tiz:


michielvandergeest:
While it's good to have extra checks in general, it seems to heavy to do these full checking run time and including a separate library to do so. Closing this PR. We may add some minimal checks where it's really needed in the future.

PR #115, opened 2024-06-20
Element tests

M4tiz:


src/engines/L3/element.js
```diff
@@ -298,23 +298,13 @@ const Element = {
...
-      for (let i = 0; i < props.length; i++) {
-        // todo: fix code duplication
-        const [p, v] = props[i]
-        if (isTransition(value)) {
-          return this.animate(p, v, value.transition)
-        } else if (this.node[p] !== v) {
-          this.node[p] = v
-        }
-      }
     }
```
michielvandergeest, line 308: I prefer leaving this unchanged, at least in this PR.

While your change works and makes sense (in fact it used to be exactly like this), from some (preliminary) tests it seems like the current code is more optimal from a JIT perspective.

In most cases there will be only 1 prop to be updated. Creating a new scope with a for loop even when there's on 1 item to loop over is then suboptimal.

We should benchmark this better though, but for now let's keep the construct as it is.

michielvandergeest (changes requested):
One comment to revert a code change. Otherwise LGTM. Thanks for contributing!

src/engines/L3/element.js
M4tiz: Reverted

michielvandergeest (approved):
LGTM. Thanks again for this contribution 👍

PR #116, opened 2024-06-24
Precompiler fix: checks objects with template property

uguraslan:
This is a quick fix for the case where config object gets template property from another object. Now the precompiler also checks objects with `template` property.

```js
const myTemplate = {
  template: '
    <Element x="620" y="20">
     <Element w="300" h="200" />
    </Element>'
}

export default Blits.Component('myComp', {
...myTemplate
})
```


michielvandergeest (approved):
LGTM

PR #117, opened 2024-06-27
Moved destroying elements & components to last

suresh-gangumalla:
- Found forEndCounter is not correctly calculated so moving destroy logic to end, after elems & components created, to make sure node destroy will happen correctly


michielvandergeest (approved):
LGTM

PR #119, opened 2024-07-02
Resolve Typescript error with both Props keys

suresh-gangumalla:
Observed typescript error while adding normal string props and advanced props in a component props key. 

Allowing only one kind of keys but not both

### DIFF
--- a/src/component.d.ts
+++ b/src/component.d.ts
@@ -91,6 +91,7 @@ declare namespace Component {
   }
   type PropsArray<T extends string> = T[]
   type AdvancedProps = AdvancedProp[]
+  type MixedProps<T extends string> = (T | AdvancedProp)[];
 
   // type ExtractPropNames<T extends Prop[]> = T extends (infer U)[]
   // ? U extends string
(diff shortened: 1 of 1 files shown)


uguraslan:
Hey @suresh-gangumalla , I tested it with the following code but I am getting an error on `props` property key:

```js
export default Blits.Component('MyComp', {
  template: `
    <Element x="400" y="100" />
    `,
  props: [
    'itemA',
    'itemB',
    {
      key: 'itemC',
      required: true,
      default: 'default value',
    },
    'itemD',
  ],
  hooks: {
    ready() {},
  },
})
```

The error: 
```
Type '(string | { key: string; required: true; default: string; })[]' is not assignable to type 'AdvancedProp[] | PropsArray<string>'.
  Type '(string | { key: string; required: true; default: string; })[]' is not assignable to type 'AdvancedProp[]'.
    Type 'string | { key: string; required: true; default: string; }' is not assignable to type 'AdvancedProp'.
      Type 'string' is not assignable to type 'AdvancedProp'.ts(2322)
component.d.ts(397, 5): The expected type comes from property 'props' which is declared here on type 'ComponentConfig<string, State, Methods>'
```

uguraslan:
@suresh-gangumalla Ignore my previous comment, there was an unrelated issue with my test setup. I tested again and the fix does not have an issue.

uguraslan (approved):
LGTM

PR #120, opened 2024-07-09
Update to Renderer RC1.0

jfboeve:
Update the Lightning Renderer package to the latest version, and applied the required changes for it to work again.

Changed the defaultFont to `sans-serif` instead of `lato`. This helps for when there are no font's imported when launching an app.

### DIFF
--- a/src/engines/L3/fontLoader.js
+++ b/src/engines/L3/fontLoader.js
@@ -1,10 +1,11 @@
-import { WebTrFontFace, SdfTrFontFace } from '@lightningjs/renderer/core'
+//import { WebTrFontFace, SdfTrFontFace } from '@lightningjs/renderer/core'
 
+import { SdfTrFontFace, WebTrFontFace } from '@lightningjs/renderer'
 import Settings from '../../settings.js'
 import { renderer } from './launch.js'
 
 export default () => {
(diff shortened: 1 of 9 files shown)


src/engines/L3/fontLoader.js
```diff
@@ -1,10 +1,11 @@
-import { WebTrFontFace, SdfTrFontFace } from '@lightningjs/renderer/core'
+//import { WebTrFontFace, SdfTrFontFace } from '@lightningjs/renderer/core'
```
michielvandergeest, line 1: let's remove this commented line

src/engines/L3/launch.js
```diff
@@ -78,8 +69,10 @@ You can remove the option from your \`src/index.js\`-file. And you can safely re
       renderer = null
     }
   }
-
-  renderer.init().then(shaderLoader).then(fontLoader).then(initApp)
+  shaderLoader()
+  fontLoader()
+  initApp()
+  // renderer.init().then(shaderLoader).then(fontLoader).then(initApp)
```
michielvandergeest, line 75: there is no need to run these in sequence (anymore)? If so, then let's removed the commented out line here as well

michielvandergeest (approved):
LGTM

PR #122, opened 2024-07-18
Precompiler changes

jfboeve:
Moved relative path line from /lib/ to /vite/, this removes the dependency to path in the /lib/ folder, which in turn allows us to use the precompiler on the playground.

Added precompiler to exports in package.json


michielvandergeest (approved):
LGTM!

PR #123, opened 2024-07-18
Remove import-chunk-url

jfboeve:
Removed import-chunk-url plugin + dependency.


michielvandergeest (approved):
LGTM. This vite plugin indeed isn't necessary anymore 👍

PR #128, opened 2024-07-30
Feat/remove create app

sandeep-vedam:
Removed create app functionality from blits and made it as a independent package


bin/index.js
```diff
@@ -1,141 +1,9 @@
...
-      },
-      () => setAppData(config),
-      () => setBlitsVersion(config),
-      () => config.esLint && addESlint(config),
-      () => config.gitInit && gitInit(config.targetDir, config.fixturesBase),
-      () => done(config)
-    ])
-  })
+try {
+  execSync('npx --yes blits-create@latest', { stdio: 'inherit' })
```
sandeep-vedam, line 5: Once after publishing to the real location we can update the location here(might be like @lightningjs/blits-create@latest

packages/blits-create/.editorconfig
```diff
@@ -0,0 +1,7 @@
+[*]
+charset = utf-8
+indent_size = 2
+indent_style = space
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
michielvandergeest, line 7: Do we need this file here? Same goes for the other files on the root level of this package, like .gitignore, but also CHANGELOG, LICENSE, etc

Since it's part of the same project, I would say that it's not needed?

michielvandergeest (changes requested):

PR #129, opened 2024-08-01
Fixes default key in for-loop for delete operation

suresh-gangumalla:
When key attribute is not provided in template then considering index as key, these keys will be used to delete previous elements.

Before this fix, keys are 'undefined' so everytime all elements are deleted and created again

### DIFF
--- a/src/lib/codegenerator/generator.js
+++ b/src/lib/codegenerator/generator.js
@@ -287,7 +287,9 @@ const generateForLoopCode = function (templateObject, parent) {
     const created${forStartCounter} = []
     const forloop${forStartCounter} = (collection = [], elms, created) => {
       const rawCollection = getRaw(collection)
-      const keys = new Set(rawCollection.map((${item}) => '' +  ${interpolate(key, '')}))
+      const keys = new Set(rawCollection.map((${item}, index) => ${
+    key ? interpolate(key, '') : 'index'
+  }))
   `)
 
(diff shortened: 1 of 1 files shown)


src/lib/codegenerator/generator.js
```diff
@@ -287,7 +287,9 @@ const generateForLoopCode = function (templateObject, parent) {
     const created${forStartCounter} = []
     const forloop${forStartCounter} = (collection = [], elms, created) => {
       const rawCollection = getRaw(collection)
-      const keys = new Set(rawCollection.map((${item}) => '' +  ${interpolate(key, '')}))
+      const keys = new Set(rawCollection.map((${item}, index) => ${
+    key ? interpolate(key, '') : 'index'
```
michielvandergeest, line 291: @suresh-gangumalla should 'index' indeed be a string? or did you mean to use the variable `index` ?

suresh-gangumalla: To use `index` as variable only

suresh-gangumalla: As per current `dev` code, when we avoid/forgot key attribute in forloop then `__index` is treated as key so the 'index' here will reference to index variable of map in generated code. 

forloop-default-key-2

michielvandergeest:
Closing as this fix has already been incorporated in another form.

PR #134, opened 2024-08-18
Optional priority-based event listener

uguraslan:
- Listeners now accept an optional numeric priority parameter (higher value = higher priority, default is 0)
- Callback functions can return `false` to stop event propagation to lower-priority listeners
- Changes maintain backward compatibility with existing listener registrations and behavior
- Priority-based sorting occurs during listener registration, using an array for efficient sorting and execution
- slightly slower registration for faster event execution

An alternative version : #135

### DIFF
--- a/src/component.d.ts
+++ b/src/component.d.ts
@@ -265,7 +265,10 @@ declare namespace Component {
     /**
     * Listen to events emitted by other components
     */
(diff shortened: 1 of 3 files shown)


michielvandergeest:
Let's move forward with the alternative approach in #135

PR #135, opened 2024-08-19
Optional priority-based event listener (sorting in the execution phase)

uguraslan:
This version is an alternative to PR #134 by changing when we sort event listeners. Instead of sorting every time we add a listener, we wait until the event actually happens. We also save the sorted (cached) list for later use. This means we only sort listeners when we really need to.

This change minimizes unnecessary sorting operations, especially for events that may never be triggered.

- Listeners now accept an optional numeric priority parameter (higher value = higher priority, default is 0)
- Callback functions can return `false` to stop event propagation to lower-priority listeners
- Changes maintain backward compatibility with existing listener registrations and behavior
- Priority-based sorting occurs during callback execution once for each event, preventing sorting overhead.
- slightly slower first event execution for fast registration and subsequent event executions

### DIFF
--- a/src/lib/eventListeners.js
+++ b/src/lib/eventListeners.js
@@ -21,13 +21,13 @@ const cache = new Map()
 export default {
   registerListener(component, event, cb, priority = 0) {
     let componentsMap = eventsMap.get(event)
-    if (!componentsMap) {
+    if (componentsMap === undefined) {
       componentsMap = new Map()
       eventsMap.set(event, componentsMap)
     }
 
     let components = componentsMap.get(component)
-    if (!components) {
+    if (components === undefined) {
       components = new Set()
       componentsMap.set(component, components)
     }
@@ -38,21 +38,24 @@ export default {
 
   executeListeners(event, params) {
     const componentsMap = eventsMap.get(event)
-    if (!componentsMap) {
-      return true // No listeners so execution can be considered successful
+    if (componentsMap === undefined || componentsMap.size === 0) {
+      return true // No listeners, so execution can be considered successful
     }
 
-    if (!cache.has(event)) {
-      // collect all callbacks and sort them by priority
+    if (cache.has(event) === false) {
       const allCallbacks = []
-      componentsMap.forEach((components, component) => {
-        components.forEach((callbackObj) => allCallbacks.push({ ...callbackObj, component }))
-      })
+      for (const [component, components] of componentsMap) {
+        for (const callbackObj of components) {
+          allCallbacks.push({ ...callbackObj, component })
+        }
+      }
       allCallbacks.sort((a, b) => b.priority - a.priority)
       cache.set(event, allCallbacks) // Cache the sorted callbacks with component context
     }
 
-    for (const { cb, component } of cache.get(event)) {
+    const callbacks = cache.get(event)
+    for (let i = 0; i < callbacks.length; i++) {
+      const { cb, component } = callbacks[i]
       const result = cb.call(component, params)
       if (result === false) {
         return false // Stop propagation if any listener returns false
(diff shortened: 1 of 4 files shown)


src/lib/eventListeners.js
```diff
@@ -31,24 +32,46 @@ export default {
...
   },
+
   executeListeners(event, params) {
     const componentsMap = eventsMap.get(event)
-    if (componentsMap) {
-      componentsMap.forEach((component) => {
-        component.forEach((cb) => {
-          cb(params)
-        })
+    if (!componentsMap) {
```
michielvandergeest, line 41: we should change this to an `undefined` check instead of a generic falsy check

src/lib/eventListeners.js
```diff
@@ -31,24 +32,46 @@ export default {
...
+      return true // No listeners so execution can be considered successful
+    }
+
+    if (!cache.has(event)) {
```
michielvandergeest, line 45: better to do a strict `=== false` check

src/lib/eventListeners.js
```diff
@@ -31,24 +32,46 @@ export default {
...
+      // collect all callbacks and sort them by priority
+      const allCallbacks = []
+      componentsMap.forEach((components, component) => {
+        components.forEach((callbackObj) => allCallbacks.push({ ...callbackObj, component }))
       })
```
michielvandergeest, line 50: this may have some performance impact - forEach inside a forEach
we should change this to a for or while loop probably and make sure that the nested loop is really necessary

src/lib/eventListeners.js
```diff
@@ -16,9 +16,10 @@
  */
 
 const eventsMap = new Map()
+const cache = new Map()
```
michielvandergeest, line 19: can we rename this to `callbackCache`? If I understood the code correctly, this map caches the callbacks., right?

src/lib/eventListeners.js
```diff
@@ -31,24 +32,46 @@ export default {
...
+    return true // All listeners executed without stopping propagation
   },
+
   removeListeners(component) {
-    eventsMap.forEach((componentMap) => {
-      const cmp = componentMap.get(component)
-      if (cmp) {
-        cmp.clear()
-        componentMap.delete(cmp)
+    eventsMap.forEach((componentsMap, event) => {
```
michielvandergeest, line 66: good moment to also refactor this `forEach` to a `while` loop

michielvandergeest (changes requested):

src/lib/eventListeners.js
uguraslan: yes, the map caches the callbacks and I agree it would be better to have non-generic variable name there. I will change it.

src/lib/eventListeners.js
uguraslan: fixed this. but I realized I need to add a check for also the size of componentsMap.

src/lib/eventListeners.js
uguraslan: fixed it.

src/lib/eventListeners.js
uguraslan: replaced with `for ... of`. I think it is faster than `forEach`. I could not get rid of the nested loop without adding intermediate arrays.

src/lib/eventListeners.js
uguraslan: renamed the variable.

src/lib/eventListeners.js
uguraslan: used `for...of` instead of `forEach`.

michielvandergeest (approved):
LGTM!

PR #137, opened 2024-08-19
Feature/performance improvements

michielvandergeest:


src/engines/L3/element.js
```diff
@@ -102,9 +101,9 @@ const propsTransformer = {
     this.props['zIndex'] = v
   },
   set color(v) {
-    if (typeof v === 'string' && v.indexOf('{') === -1) {
+    if (typeof v === 'string' && v.startsWith('{') === false) {
```
wouterlucas, line 104: nice one ❤️

src/engines/L3/element.js
```diff
@@ -266,65 +258,63 @@ const Element = {
...
+    if (this.props.props['color'] === undefined && '__textnode' in props === false) {
       this.props.props['color'] = 0
     }
 
     this.node = props.__textnode
       ? renderer.createTextNode({ ...textDefaults, ...this.props.props })
       : renderer.createNode(this.props.props)
 
-    if (props['@loaded']) {
+    if ('@loaded' in props === true && typeof props['@loaded'] === 'function') {
```
wouterlucas, line 269: 👀 

is `'@loaded' in props === true` faster then an undefined check?

src/lib/codegenerator/generator.js
```diff
@@ -287,7 +306,12 @@ const generateForLoopCode = function (templateObject, parent) {
     const created${forStartCounter} = []
     const forloop${forStartCounter} = (collection = [], elms, created) => {
       const rawCollection = getRaw(collection)
-      const keys = new Set(rawCollection.map((${item}) => '' +  ${interpolate(key, '')}))
+      const keys = new Set()
+      let l = rawCollection.length
+      while(l--) {
```
wouterlucas, line 311: ❤️  while loops for the win

src/lib/colors/colors.js
```diff
@@ -39,17 +39,16 @@ export default {
...
     if (color.startsWith('0x') && color.length === 10) {
-      colorsMap[color] = color
       return color
     }
 
+    // it is a previously mapped color, return it
+    if (colorsMap[color] !== undefined) {
+      return colorsMap[color]
+    }
+
```
wouterlucas, line 51: this means you're no longer caching `0x` hex colors, is `(color.startsWith('0x') && color.length === 10)` faster then `(colorsMap[color] !== undefined)`?

wouterlucas (approved):
LGTM just some Qs for learnings 👍

src/lib/colors/colors.js
michielvandergeest: Correct, the difference between the 2 check is close to negligible. In the case of `0x` colors it doesn't really make sense to cache the result as the same value, without manipulation, will be returned anyway. So with this change we don't store data unnecessarily. For most normal app scenarios this won't be a problem, but when working with random colors in test scenarios the cache can fill up.

src/engines/L3/element.js
michielvandergeest: No it's not :smile: .. changing to `undefined` check

src/lib/colors/colors.js
wouterlucas: Fair enough - thanks for explaining!

PR #141, opened 2024-08-22
Update Component's method names to standard format

suresh-gangumalla:
1. Updated focus, select and trigger method names by prefixing with '$'
2. Added deprecation log for previous signature of methods


michielvandergeest (approved):
LGTM

PR #144, opened 2024-08-26
Global appState plugin

michielvandergeest:


suresh-gangumalla:
Verified following scenarios

- `appState` property as dynamic variable
- `appState` property as reactive variable
- `appState` property in computed property

PR #145, opened 2024-08-26
Moved  $log functionality to a plugin (loaded by default).

michielvandergeest:
Moved $log functionality to a plugin.

This solves an issue where the App logger is instantiated before settings are present, which makes the debugLevel setting have no influence on logs specified in the App code.


suresh-gangumalla:
Changes are perfectly working and resolved the issue reported in discord

suresh-gangumalla (approved):

PR #146, opened 2024-08-27
`.blits` file type support

uguraslan:
- Implemented a new vite plugin to handle & transform `.blits` files. It transforms them into existing Javascript format so it works even precompile is on or off.
- Updated precompiler to limit it to process only `.js` or `.ts` files
- Added a generic type definition `blits.d.ts` to supress editors' type warnings where a `.blits` file is imported into a JavaScript or TypeScript file.
- There must be only one component call in each `.blits` file. 

These changes are backward compatible and do not break anything.

#### Additional info about the file format:

- The file may begin with a JavaScript-style block comment (`/* ... */`), typically used for license texts or similar content.
- Template strings must be enclosed within `<template>` and `</template>` tags.
- JavaScript/TypeScript content must be enclosed within `<script>` and `</script>` tags.
- The <script> tag can include a `lang="ts"` attribute to specify TypeScript as the content language. JavaScript is the default.
- Each `.blits` file can contain only one Blits Component.
- JavaScript or TypeScript components can import `.blits` components, provided the file extension is included in the import path.
- `.blits` components can also import JavaScript/TypeScript components using standard import syntax.

Example:

```html
<template>

    <Element>
      <!-- simple square component that takes a size (number) argument and maps it to w and h -->
      <Square x="100" y="100" size="50" />
      <Square x="100" y="200" size="100" />
      <Square x="100" y="350" size="200" />
      <!-- reactive (animated) x position for component -->
      <Square :x.transition="$x" y="600" size="50" />
      <!-- card component that takes a string size argument and also has a nested square component -->
      <Card x="500" y="100" />
      <Card x="500" y="500" size="large" />

      <Element x="1000" y="100">
        <Text>Dynamic components</Text>
        <Component is="$dynamicComponent" y="100" color="blue" />
        <Component :for="(bla, index) in $dynamicComponents" is="$bla" :x="205 * $index" y="300" />
      </Element>
    </Element>

</template>

<script>
import Blits from '@lightningjs/blits'
import Square from '../components/Square.blits'
import Card from '../components/Card' // importing JavaScript component
import Button from '../components/Button.blits'

export default Blits.Component('Components', {
  components: {
    Square,
    Card,
  },
  state() {
    return {
      x: 100,
      dynamicComponent: Button,
      dynamicComponents: ['Square', 'Card', 'Square', Button],
      toggle: false,
      size: 50,
    }
  },
  hooks: {
    ready() {
      this.$setInterval(() => {
        this.x = this.x === 100 ? 250 : 100
        this.size = this.size === 50 ? 200 : 50
      }, 2000)
    },
  },
})
</script>
```

### DIFF
--- a/boilerplate/js/default/blits.d.ts
+++ b/boilerplate/js/default/blits.d.ts
@@ -1,0 +1,5 @@
+// required for suppressing vscode typescript errors when importing .blits files in JS/TS files
+declare module '*.blits' {
+  const value: any;
+  export default value;
+}
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "./vite": "./vite/index.js",
     "./transitions": "./src/router/transitions/index.js",
     "./precompiler": "./src/lib/precompiler/precompiler.js",
+    "./blitsFileConverter": "./src/lib/blitsfileconverter/blitsfileconverter.js",
     "./plugins": "./src/plugins/index.js"
   },
   "scripts": {
--- a/src/lib/blitsfileconverter/blitsfileconverter.js
+++ b/src/lib/blitsfileconverter/blitsfileconverter.js
@@ -15,28 +15,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export default function blitsFileType() {
-  return {
-    name: 'vite-plugin-blits-file-type',
-    enforce: 'pre',
-    transform(src, id) {
-      if (id.endsWith('.blits')) {
-        try {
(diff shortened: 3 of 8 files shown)


jfboeve:
Can you move the added functions for parsing .blits files to lib similar to the precompiler? Please also add a export to the package.json. This way I can use these functions for the Playground.

uguraslan:
Hey @jfboeve, good idea. I will do that.

michielvandergeest (approved):
LGTM! 🚀

PR #147, opened 2024-08-28
support nested prop watch

suresh-gangumalla:
This PR handles watching on nested property of state and global appState.

```
state(){
 return {
    foo: {
      bar: 1
    }
 }
},
watch: {
  foo(v) {
    // first level properties can be directly used as method names
  }, 
  'foo.bar': function(v) {
    // nested propeties can be watched like this format
  }
}

```


michielvandergeest (approved):
LGTM!

PR #149, opened 2024-08-28
Fixes index reference in key attribute

suresh-gangumalla:
if index is used in key attribute then this fix make sure corresponding 'index' variable is available within while loop scope.

template:
```
 <Element :for="(item, index) in $data" key="$index + 'key'" > 
```
Generated code: 
```
  while(l--) {
    const item = rawCollection[l]
    const index = l
    keys.add('' +   index + 'key')
 }
```

### DIFF
--- a/src/lib/codegenerator/generator.js
+++ b/src/lib/codegenerator/generator.js
@@ -310,6 +310,7 @@ const generateForLoopCode = function (templateObject, parent) {
       let l = rawCollection.length
       while(l--) {
         const ${item} = rawCollection[l]
+        const ${index} = l
         keys.add('' +  ${interpolate(key, '') || 'l'})
       }
   `)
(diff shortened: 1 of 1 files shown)


michielvandergeest:
But the index variable shouldn't be used in the key attribute. That defeats the point of using a key to begin with ...

https://lightningjs.io/v3-docs/blits/built-in/for-loop.html#the-importance-of-using-the-key-attribute

> Also beware that we can't rely on the index parameter provided in for loop, because that only indentifies the position in the Array, and not the actual item itself.

suresh-gangumalla:
This PR deals with worst case scenario where developer uses `index` variable in `key` attribute, instead of breaking, the forloop execution now it's proceeds.   

Added console warning message to notify the user that using `index` in `key` attribute is highly not recommended.

uguraslan (approved):
LGTM

PR #150, opened 2024-08-30
Fix use of multiple instances of same plugin

jfboeve:
This fix allows us to use multiple instances of one plugin when using a specified name.


michielvandergeest (approved):
LGTM!

PR #154, opened 2024-09-09
Normalize colors in effect

suresh-gangumalla:


src/engines/L3/element.js
```diff
@@ -205,6 +205,11 @@ const propsTransformer = {
     }
   },
   set effects(v) {
+    for (let i = 0; i < v.length; i++) {
+      if (v[i].props && v[i].props.color) {
```
michielvandergeest, line 209: maybe this should be a `'props' in v[i] === true` check (and `'color' in v[i].props` as well). I think that's (slightly) faster

PR #161, opened 2024-09-18
Added explicit ImageTexture constructor name check

suresh-gangumalla:
Static checking of constructor name as ```_ImageTexture``` leading to issues in build version due to minification so checking directly with ```ImageTexture```

### DIFF
--- a/src/lib/reactivity/reactive.js
+++ b/src/lib/reactivity/reactive.js
@@ -17,6 +17,7 @@
 
 import { track, trigger, pauseTracking, resumeTracking } from './effect.js'
 import symbols from '../symbols.js'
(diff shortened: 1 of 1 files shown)


src/lib/reactivity/reactive.js
```diff
@@ -17,6 +17,7 @@
 
 import { track, trigger, pauseTracking, resumeTracking } from './effect.js'
 import symbols from '../symbols.js'
+import { ImageTexture } from '@lightningjs/renderer'
```
michielvandergeest, line 20: can we move the import from an external package to the top?

michielvandergeest (approved):
LGTM

PR #165, opened 2024-09-20
Percentage calculation

michielvandergeest:


suresh-gangumalla:
Changes are working as expected.

PR #167, opened 2024-09-25
Feat/config charset

suresh-gangumalla:


vite/msdfGenerator.js
```diff
@@ -158,6 +163,44 @@ const copyDir = (src, dest) => {
...
+  // Check if font config.json is exists
+  if (fs.existsSync(configFilePath)) {
+    if (!fs.existsSync(charsetDir)) {
+      fs.mkdirSync(charsetDir, { recursive: true })
+    }
+    const configData = JSON.parse(fs.readFileSync(configFilePath, 'utf8'))
+    let charsetContent = configData.charset
+    const presets = configData.preset
+    for (let i = 0; i < presets.length; i++) {
+      charsetContent = charsetContent + presets[i]
```
michielvandergeest, line 192: this would just add the preset name to the charset, instead of appending the characters associated with the preset, right?

suresh-gangumalla:
Dependency on [CharsetFilepath Optional argument](https://github.com/lightning-js/msdf-generator/pull/6)

vite/msdfGenerator.js
```diff
@@ -158,6 +163,44 @@ const copyDir = (src, dest) => {
...
+
     entry.isDirectory() ? copyDir(srcPath, destPath) : fs.copyFileSync(srcPath, destPath) // Copy files
   }
 }
+
+const createCharsetFile = (charsetDir, ConfigDir, fontName) => {
+  const charsetFilePath = path.join(charsetDir, `${fontName}.charset.txt`)
+
+  // Check charset file already exists, if exists, return file path
+  if (fs.existsSync(charsetFilePath)) {
```
michielvandergeest, line 177: This check means that making changes to the charset config will never be reflected

vite/msdfGenerator.js
```diff
@@ -158,6 +163,44 @@ const copyDir = (src, dest) => {
...
+    const configData = JSON.parse(fs.readFileSync(configFilePath, 'utf8'))
+    let charsetContent = configData.charset
+    const presets = configData.preset
+    for (let i = 0; i < presets.length; i++) {
+      charsetContent = charsetContent + presets[i]
+    }
+
+    // write charset content into .txt file
+    try {
+      fs.writeFileSync(charsetFilePath, charsetContent)
```
michielvandergeest, line 197: is the .txt file added to the public/assets folder? or the 'hidden' folder in node_modules?

michielvandergeest (changes requested):

vite/msdfGenerator.js
suresh-gangumalla: No, It will append all characters in preset array to charset

vite/msdfGenerator.js
suresh-gangumalla: Before that we have one more check, ```msdf.json``` or ```.msdf.png```  available in ```.tmp-msdf-fonts-v2``` directory or not.  if not exist then only we are going to generate them.

I think, User must delete ```.tmp-msdf-fonts-v2``` directory to generate again.

vite/msdfGenerator.js
suresh-gangumalla: it will be part of ```.tmp-msdf-fonts-v2``` directory and ```.txt``` file won't be copied to ```dist``` directory

vite/msdfGenerator.js
suresh-gangumalla: Wait, actually locally I have different logic let me check again.

vite/msdfGenerator.js
michielvandergeest: okay, yeah that's an interesting point - cause I can imagine a user adding something to the config later on. that should ideally lead to an automatic regeneration

vite/msdfGenerator.js
suresh-gangumalla: This comment no longer required as charset text file generation completely removed from vite plugin.

vite/msdfGenerator.js
michielvandergeest: I had an idea for solving this: instead of looking if the file already exists in `.tmp-msdf-fonts-v2`, can't we compare file dates? If the font file or the config file (in the `assets` folder) are newer than those generated in the `.tmp-msdf-fonts-v2`-folder, something must have changed so we generate again

This removes the need for a user to ever manually delete `.tmp-msdf-fonts-v2` to generate again ...

Does this make sense?

cc @uguraslan

uguraslan: I suggest we save the checksum of any config files associated with each font into the `tmp-msdf-fonts-v2` directory. For each font file, the process would be:

- Check if the font has an associated config file.
- If a config exists, check if a previously calculated checksum is saved.
  - If no checksum is found, regenerate the font and save the new checksum.
  - If a checksum is found, compare it with the current checksum. If they differ, regenerate the font and update the checksum.
- If no config file exists, check if the font has already been generated.
  - If it hasn't been generated, create it.
  - If it has, do nothing.
  
For the checksum calculation, we can use a fast algorithm like `md5`, as we don't need cryptographic security—just a basic change detection. The checksums could be stored in files like `.tmp-msdf-fonts-v2/font-name.msdf.checksum` for each font with a config file.

suresh-gangumalla: As per your suggestion @uguraslan , we are going to generate checksum for config.json not for .ttf file.  As ```font.config.json``` would be small size, I feel, generating checksum everytime in ```dev``` mode does not alter performance that can be noticed.  

There are very less chances that the ```.ttf``` file get modified but still, I feel, we should address that as well but creating checksum for ```ttf``` file could be more intensive task. 

I feel, reading file metadata is less costly here for ```ttf``` files. like modified time

vite/msdfGenerator.js
suresh-gangumalla: Adding presets defintion to charset logic moved to msdf-generator so this comment can be out of blits vite plugin scope.  Resolving this conversation.

uguraslan:
@suresh-gangumalla I have tested the PR and it works as expected.

But we need to upgrade the version of msdf-generator to `1.1.0` in order this changes to work. So we need a package.json/package-lock.json update.

Also there are still old comments and variable names with "ttf". They don't affect anything right now. But it would be great to update them also while updating `package.json`

uguraslan (approved):
LGTM

PR #171, opened 2024-09-27
Added fix for interupting running / scheduled transitions

### LINKED ISSUE #170: Transition skips to end when interrupted
I think for Transitions to be useful, they should be interruptible without skipping to the end (whether by default or through some property). 

In the case of a user scrolling a row of tiles, a quick right then left press you would expect the row to start scrolling right, stop where it is when the user presses left, then go back to where it came from. You would not what it to jump to the end of the initial right-directed transition.

Quickly pressing right/left in the example below shows what I mean. 

https://playground.lightningjs.io/#2Ass8WOF4aJM

https://github.com/user-attachments/assets/24d1f24a-260b-440e-9e61-e09917001566

Just noting the behavior seems a bit consistent as well, I've sometimes seen the transition not jump to the end when interrupted.

michielvandergeest:
Fixes #170.

Depends on https://github.com/lightning-js/renderer/pull/401

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -346,13 +346,13 @@ const Element = {
   animate(prop, value, transition) {
     // if current value is the same as the value to animate to, instantly resolve
     if (this.node[prop] === value) return
-    // check if a transition is already scheduled or running on the same prop
-    if (this.scheduledTransitions[prop]) {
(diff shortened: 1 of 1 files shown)


wouterlucas (approved):
LGTM

suresh-gangumalla:
Changes are working as expected.

suresh-gangumalla (approved):
LGTM

PR #173, opened 2024-10-01
Update launch.js, remove SDF for canvas rendering mode

promentol:
If render engine is canvas, SDF text rendering engine is not available, thus its going to throw error in future processing


michielvandergeest (approved):
LGTM

PR #175, opened 2024-10-01
Release 1.5.0

michielvandergeest:
- Fixed issue with interupting running / scheduled transitions with a new transition on the same prop
- Fixed issue with (incompatible) SDF text renderer being loaded when `renderMode` is set to convas
- Added support for handling key release (i.e `keyup`) events


wouterlucas (approved):
LGTM

PR #184, opened 2024-10-09
Fix/cleanup effects global reactivity

michielvandergeest:


suresh-gangumalla (approved):
Implemented logic is working as expected. Verified Blits example application Theming & Translation sample where plugins  comes to action. 

Good to Merge

PR #187, opened 2024-10-14
adds .otf and .woff support to the vite plugin

uguraslan:
`msdf-generator` supports `.otf` and `.woff` font type. I have updated the msdf plugin to also monitor those font file types in Blits projects.


michielvandergeest (approved):
LGTM

PR #193, opened 2024-10-21
Feature/ts rewrite

michielvandergeest:


index.d.ts
```diff
@@ -15,19 +15,802 @@
...
+
+  interface Transition {
+    /**
+     * Name of the prop to transition (i.e. 'x', 'y', 'alpha', 'color')
+     */
+    prop: string,
+    /**
+     * Value the prop should transition to (i.e. 0, 100, '#223388')
+     */
+    value: any,
```
wouterlucas, line 372: are these typeable as well ? e.g. number | string | boolean | CustomType?

index.d.ts
```diff
@@ -15,19 +15,802 @@
...
+
+  interface Before {
+    /**
+    * Name of the prop to set before the transition starts
+    */
+    prop: string,
+    /**
+     * Value the prop to set before the transition starts
+     */
+    value: any,
```
wouterlucas, line 395: same here, can we add a type here?

index.d.ts
```diff
@@ -15,19 +15,802 @@
...
+    /**
+     * Message to be announced
+     */
+    message: String,
+    /**
+     * Politeness level
+     *
+     * Defaults to 'off'
+     */
+    politeness?: 'off' | 'polite' | 'assertive'
```
wouterlucas, line 425: what does this do?

index.d.ts
```diff
@@ -15,19 +15,802 @@
...
+    */
+    type: 'msdf' | 'sdf',
+  } & (SdfFontWithFile | SdfFontWithPngJson)
+
+  type Font = WebFont | SdfFont
+
+  type ShaderEffect = {
+    name: string,
+    type: RendererShaderEffect
+  }
```
wouterlucas, line 538: I guess this works for now - but we should think about making this easier when the new shader / effect refactor from Jeff comes in :)

wouterlucas (approved):
LGTM!

suresh-gangumalla (approved):
LGTM

index.d.ts
michielvandergeest: it's a property used in text to speech to indicate critical notifications that require immediate attention

index.d.ts
michielvandergeest: yes, this is temporary for now, and up for general refactor

PR #197, opened 2024-10-29
Added support for multiple language files

suresh-gangumalla:
Added support to include separate langugage file
for each language code in translation.json


michielvandergeest (approved):
LGTM

PR #201, opened 2024-10-31
Fix: Export Interfaces and Change Return Type for beforeHooks Function

vonvick:
#### What does this PR do?
This PR exports various interfaces in the Blits namespace and extended the return types for the router Hooks `before` function.


michielvandergeest (approved):
LGTM

PR #210, opened 2024-11-06
Added support for fav icon in browser

sandeep-vedam:
Added support for favicon when user launches app in browser


packages/create-blits/boilerplate/js/default/index.html
```diff
@@ -3,6 +3,7 @@
       <style>
         html, body, * { padding: 0; margin: 0}
       </style>
+      <link rel="icon" href="public/assets/favicon.svg" type="image/svg">
```
suresh-gangumalla, line 6: Verify public directory path in production dist.

sandeep-vedam: I have verified in production and there is no issue. Here the svg file is copied into assets inside dist folder and in index file the path is updated without the public folder and its like /assets/favicon.svg

michielvandergeest (approved):

PR #211, opened 2024-11-07
Added passing all props in a single object

suresh-gangumalla:
Updated to accept all props in a single object under the `props` key

Supports both reactive and dynamic


suresh-gangumalla:
Closing this PR as passing all props in a single object introduces unnecessary complexity and performance overhead in reactive mode

PR #213, opened 2024-11-08
Fix re focussing same component issue

suresh-gangumalla:
Addressess issue with accindental re-foucsing same component again


michielvandergeest (approved):
LGTM!

PR #218, opened 2024-11-15
Fixes cached router view holder's alpha issue

suresh-gangumalla:
with default transition, alpha is set to zero for previous route view holder element. if preivous route is configured to keep alive then that view is cached with holder alpha set to zero

Now, if previous route has a custom transition and if it does not contain alpha related prop transition in 'before' or 'in' then holder's alpha will not get updated to 1 from zero set by previous route out transition

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -170,23 +170,23 @@ export const navigate = async function () {
         holder = view[symbols.holder]
 
         // Check, whether cached view holder's alpha prop is exists in transition or not
-        let isAlphaPropExists = false
+        let hasAlphaProp = false
         if (route.transition.before) {
           if (Array.isArray(route.transition.before)) {
             for (let i = 0; i < route.transition.before.length; i++) {
               if (route.transition.before[i].prop === 'alpha') {
-                isAlphaPropExists = true
+                hasAlphaProp = true
                 break
               }
             }
           } else {
             if (route.transition.before.prop === 'alpha') {
-              isAlphaPropExists = true
+              hasAlphaProp = true
             }
           }
         }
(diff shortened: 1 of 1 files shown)


jfboeve (commented):
I recommend changing the variable name `isAlphaPropExists` -> `hasAlphaProp`

michielvandergeest:
> I recommend changing the variable name `isAlphaPropExists` -> `hasAlphaProp`

yes I agree. @suresh-gangumalla can you make the change accordingly?

src/router/router.js
```diff
@@ -168,6 +168,27 @@ export const navigate = async function () {
...
+              }
+            }
+          } else {
+            if (route.transition.before.prop === 'alpha') {
+              hasAlphaProp = true
+            }
+          }
+        }
+        // set holder alpha when alpha prop is not exists in route transition
+        if (!hasAlphaProp) {
```
michielvandergeest, line 189: please do a `false` check instead of a `falsey` check here

src/router/router.js
```diff
@@ -168,6 +168,27 @@ export const navigate = async function () {
...
+          if (Array.isArray(route.transition.before)) {
+            for (let i = 0; i < route.transition.before.length; i++) {
+              if (route.transition.before[i].prop === 'alpha') {
+                hasAlphaProp = true
+                break
+              }
+            }
+          } else {
+            if (route.transition.before.prop === 'alpha') {
+              hasAlphaProp = true
```
michielvandergeest, line 184: this could be combined in an `else if`?

michielvandergeest (changes requested):

michielvandergeest (approved):
LGTM

PR #225, opened 2024-11-25
Fixes Router Focused Component hasFocus

suresh-gangumalla:
Avoided directly invoking Focus.set(view/previous-focused-comp) approach which actually missing setting hasFocus flag on its component to True.

Now, Invoking $focus on target view or previously focused component from cache so that it will make sure hasFocus correctly set


michielvandergeest (approved):
LGTM

PR #228, opened 2024-11-27
Added Router BeforeAll Hook

suresh-gangumalla:
- Updated Router configuration to include router hooks along with routes from Application instance
- Executing router beforeAll hook on every route navigation

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -424,7 +424,7 @@ declare module '@lightningjs/blits' {
   }
 
   export interface RouterHooks {
-    beforeAll?: (to: Route, from: Route) => string | Route | Promise<string | Route>;
+    beforeEach?: (to: Route, from: Route) => string | Route | Promise<string | Route>;
   }
 
   export interface RouterConfig {
@@ -452,11 +452,15 @@ declare module '@lightningjs/blits' {
   export interface ApplicationConfig<P extends Props, S, M, C, W> extends ComponentConfig<P, S, M, C, W> {
     /**
      * Router configuration
+     *
+     * Configure either router or routes; if both are provided, router takes precedence
      */
     router?: RouterConfig
 
     /**
      * Routes definition
+     *
+     * Configure either router or routes; if both are provided, router takes precedence
      *
      * @example
      *
--- a/index.d.ts
+++ b/index.d.ts
@@ -449,31 +449,33 @@ declare module '@lightningjs/blits' {
     routes?: Route[]
   }
 
-  export interface ApplicationConfig<P extends Props, S, M, C, W> extends ComponentConfig<P, S, M, C, W> {
-    /**
-     * Router configuration
-     *
-     * Configure either router or routes; if both are provided, router takes precedence
+  export type ApplicationConfig<P extends Props, S, M, C, W> = ComponentConfig<P, S, M, C, W> & (
+    {
+      /**
+       * Router Configuration
+       */
+      router?: RouterConfig,
+      routes?: never
+    }
+    |
+    {
+      router?: never
+      /**
+       * Routes definition
+       *
+       * @example
+       *
+       * ```js
+       * routes: [
+       *  { path: '/', component: Home },
+       *  { path: '/details', component: Details },
+       *  { path: '/account', component: Account },
+       * ]
+       * ```
      */
-    router?: RouterConfig
-
-    /**
-     * Routes definition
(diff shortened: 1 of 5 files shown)


michielvandergeest (changes requested):
This code looks good, but as far as I understand it's now only possible to provide a router object in the ApplicationConfig

I think ideally we would keep the 'simple' configuration with direct routes (using the `routes`-property, along side the advanced configuration using the `router`-key.

If possible, the Typescript definitions should indicate that only 1 configuration (`routes` vs `router`) is allowed at the same time

index.d.ts
```diff
@@ -373,7 +373,16 @@ declare module '@lightningjs/blits' {
     watch?: W & ComponentContext<P, S, M, C>
   }
 
-  export interface ApplicationConfig<P extends Props, S, M, C, W> extends ComponentConfig<P, S, M, C, W> {
+  export interface RouterHooks {
+    beforeAll?: (to: Route, from: Route) => string | Route | Promise<string | Route>;
```
michielvandergeest, line 377: let's name this `beforeEach` instead of `beforeAll`

index.d.ts
```diff
@@ -390,6 +399,27 @@ declare module '@lightningjs/blits' {
     routes?: Route[]
   }
 
+  export interface ApplicationConfig<P extends Props, S, M, C, W> extends ComponentConfig<P, S, M, C, W> {
+    /**
+     * Router configuration
```
michielvandergeest, line 404: we should mention that router config take precedence over routes - I guess there's no way to specifiy in typescript that we should either use router or routes but not both? @jfboeve you any idea on this?

index.d.ts
```diff
@@ -390,6 +399,27 @@ declare module '@lightningjs/blits' {
...
+     */
+    router?: RouterConfig
+
+    /**
+     * Routes definition
+     *
+     * @example
```
michielvandergeest, line 411: mention here as well that router takes precedence

src/component/setup/index.js
```diff
@@ -53,8 +53,8 @@ export default function (component, config) {
   // // setup watchers
   if (config.watch) setupWatch(component, config.watch)
 
-  // // setup routes
-  if (config.routes) setupRoutes(component, config.routes)
+  // // setup router
+  if (config.router || config.routes) setupRouter(component, config.router || config.routes)
```
michielvandergeest, line 57: we can make an undefined check out of these

michielvandergeest (changes requested):

index.d.ts
jfboeve: hey sorry for the late response. got a snippet for you [here](https://www.typescriptlang.org/play/?#code/FASwdgLgpgTgZgQwMZQAQGED2BbADpsKSLMOEAc1QG9hgBfW8aeZNAJUwFdmSzKbUqABaZMAawDOALlQByEeImyANMEEwu0aagkQY4cgG0AuvUaRYiFBgJ9qa1Bu6wZszACMAVlCQRZDpy0ZXX0wI1MGYAgAT1w0AEFcXAAbECQECBACXgpUAF4bPAIiCBzKADJUAAoHAUFHTRdUDmcYMtV6hucJAH4ZQgA3WAcGAB9ah3VGmD7UQeHOwKhtEIMTEeAASlokAl1UBCSymUSUtIyssDL8+ynWmTrBBUlXZ6UO+qXtQ3kEZOTMLJTIIGJFgLswPtDrgygAmE5JVLpTLZWy5Ap1L4yH5CP4AlRyCAAd0BER2ewgByOaPIAGYEWdkZdrhiAo1vr9-oDlISSUCPksYA9JsJRC85G8VCKsagcXjSRs6KhgEA)

src/component/setup/index.js
```diff
@@ -53,8 +53,9 @@ export default function (component, config) {
   // // setup watchers
   if (config.watch) setupWatch(component, config.watch)
 
-  // // setup routes
-  if (config.routes) setupRoutes(component, config.routes)
+  // // setup router
+  const routerConfig = config.router !== undefined ? config.router : config.routes !== undefined ? config.routes : undefined
```
michielvandergeest, line 57: is this part `config.routes !== undefined ? config.routes : undefined` really needed here? 🤔

michielvandergeest (approved):
LGTM

PR #230, opened 2024-11-28
Added support to display Blits and Renderer version in inspector log

sandeep-vedam:
Added support to display Blits and Renderer version in browser  inspector console log


michielvandergeest (approved):
LGTM

PR #231, opened 2024-11-28
Added float attribute on element

suresh-gangumalla:
Introduced float attribute to position element at center or right

### DIFF
--- a/docs/essentials/element_attributes.md
+++ b/docs/essentials/element_attributes.md
@@ -36,6 +36,37 @@ Besides using values in pixels (i.e. `w="100" h="300"`), you can also specify _p
 The percentage value specified for `w` and `x` will be calculated as the percentage of the _width_ (`w`) of the parent element.
 And the percentage specified for `h` and `y` will use the _height_ (`h`) of the parent element as the base of the percentage calculation.
 
+### Predefined placement options
+
+In addition to the absolute positions that Blits and Lightning use, there are a few predefined placement options available. By adding the `placement`-attribute to an Element we can easily _center_ or align an Element to a corner, without adding custom calculations.
+
+On the `x`-axis the following _placement_ options can be used: `left` (default), `center` and `right`. On the `y`-axis the _placement_ attribute accepts `top` (default), `middle` and `bottom`.
+
+The _placement_ attribute also accepts and object with an `x` and a `y` key, to specify a placement for both axes.
+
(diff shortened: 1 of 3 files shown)


michielvandergeest:
Thanks for the work so far, a few comments / additional requirements:

- we initially choose `float` as an attribute name (to avoid `align` and `position`), but let's use `placement` instead
- the current implementation only allows for horizontal placement, we should also consider vertical

I think we should support the following use cases:

- `<Element w="100" h="100 placement="center" />` (x and y center)
- `<Element w="100" h="100 placement="right" />` (x right, y top)
- `<Element w="100" h="100 placement="{x: 'center', y: 'bottom'}" />` (x center, y bottom)

docs/essentials/element_attributes.md
```diff
@@ -36,6 +36,37 @@ Besides using values in pixels (i.e. `w="100" h="300"`), you can also specify _p
 The percentage value specified for `w` and `x` will be calculated as the percentage of the _width_ (`w`) of the parent element.
 And the percentage specified for `h` and `y` will use the _height_ (`h`) of the parent element as the base of the percentage calculation.
 
+### Predefined placement options
+
+In addition to the absolute positions that Blits and Lightning use, there are a few predefined placement options available. By adding the `placement`-attribute to an Element we can easily _center_ or align an Element to a corner, without adding custom calculations.
```
jfboeve, line 41: Would change the second sentence to: By add the `placement`-attribute to an Element we can easily align it to the center, left, right, top, bottom, or even a combination like `{y: 'center', x: 'left'}` for example without having to calculate the positions yourself.

docs/essentials/element_attributes.md
```diff
@@ -36,6 +36,37 @@ Besides using values in pixels (i.e. `w="100" h="300"`), you can also specify _p
...
+  <Element w="40" h="40" x="54" placement="bottom" color="#be185d" />
+
+  <!-- x/y placement -->
+  <Element w="40" h="40" placement="{x: 'center', y: 'bottom'}" color="#a3e635" />
+  <Element w="40" h="40" placement="{x: 'right', y: 'middle'}" color="#65a30d" />
+  <Element w="40" h="40" placement="{x: 'center', y: 'middle'}" color="#4d7c0f" />
+</Element>
+```
+
+> note: the `x` or `y` attribute is ignored for those axes defined in the `placement`-attribute
```
jfboeve, line 68: note: the `x` or `y` attributes are ignored when used in the `placement`-attribute.

PR #233, opened 2024-12-03
Fixes show attribute issue when w & h not exists

suresh-gangumalla:
When width and height is not configured on an Element then observed show attribute `true` value is setting undefined as value. This will be an issue for Elements without w & h attributes

setting w & h to zero when those attributes not configured on that particular Elmeent tag


michielvandergeest (approved):
LGTM

PR #238, opened 2024-12-06
Fix background color when disabling Render-to-Texture (RTT) property

erikhaandrikman:
This update addresses an issue with the rtt (render-to-texture) property.

Problem: When rendering a node to a texture, the color is set to 0xffffffff to correctly display during the fragmentation step. However, if a user disables the rtt property, the background color remains white (0xffffffff), resulting in a white background on the node.

Fix: When the rtt property is disabled, the color is set to 0x00000000 (transparent alpha channel),  preventing the white background issue.

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -251,8 +251,8 @@ const propsTransformer = {
   },
   set rtt(v) {
     this.props['rtt'] = v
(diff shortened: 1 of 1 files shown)


michielvandergeest (approved):
LGTM!

PR #241, opened 2024-12-10
Feat: Adding the ability to deregister event listeners

AndresMonk:
Adds the ability to remove listeners that have been registered.

### DIFF
--- a/docs/components/utility_methods.md
+++ b/docs/components/utility_methods.md
@@ -177,6 +177,42 @@ export default Blits.Component('MyComponent', {
 })
 ```
 
+### $removeListener
+
+The `$removeListener` utility method (which is available on each Blits Component) is designed to remove event listeners that were previously registered using `$listen()`. This method helps in managing event listeners manually when needed.
+
+The first argument of the `$removeListener()` method is the event `name` to stop listening for. This ensures that all listeners for the specified event are removed from the component.
+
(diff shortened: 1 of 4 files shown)


michielvandergeest:
Great work @AndresMonk! Love contributions with tests _and_ documentation!

About the implementation:he `$removeListener` method seems removes events by event type, rather than a single listener by id. Also, if I understood correctly this method will remove all listeners that exists for a specific event. I imagine that beign useful in certain cases, but I can also see how more granular control of which listeners are being removed may come in handy. What do you think?

AndresMonk:
Thanks @michielvandergeest !

_Also, if I understood correctly this method will remove all listeners that exists for a specific event._
That wasn’t what I was gunning for, so it could be my understanding of the maps is a bit wonky.


As a breakdown of how I thought this was working:

```
//! Get all the components that are listening to this specific event.
let componentsMap = eventsMap.get(event)

...

//! Check if the component exists within the map the event.
if (componentsMap.contains(component)) {

  //! If it does exist, remove this component from the listeners for this event.
  componentsMap.delete(component);

  //! Set the updated list of components for the map.
  eventsMap.set(event, componentsMap)
  callbackCache.delete(event)
}
```

michielvandergeest:
No, I think you are actually right! I looked at the implementation to fast :)

michielvandergeest (approved):
Thanks for making these last changes. This looks good to me!

PR #243, opened 2024-12-12
Remove transition end callbacks on destroy

suresh-gangumalla:


michielvandergeest:
@suresh-gangumalla I've pushed a change which I think will make the solution more optimized by stopping the animations upon destroy. Can you verify if the fix still works with this change included?

Other than that it looks all good to me.

src/engines/L3/element.js
```diff
@@ -540,12 +541,18 @@ const Element = {
...
+        this.scheduledTransitions[prop].transition === undefined
+      ) {
+        return
+      }
       // fire transition end callback when animation ends (if specified)
       if (transition.end && typeof transition.end === 'function') {
         transition.end.call(this.component, this, prop, this.node[prop])
       }
+      // remove the prop from scheduled transitions
+      this.scheduledTransitions[prop] = undefined
```
suresh-gangumalla, line 555: @michielvandergeest  I feel, instead of assinging undefined, deleting prop from scheduled transition would reduce the count of transition props so that for-loop over already stopped animation prop is not required.

michielvandergeest: Yes, that's a good point. Let's change it to deleting the prop

michielvandergeest (approved):

PR #244, opened 2024-12-12
TypeScript support for `.blits` files

uguraslan:
- Added TypeScript transpiling support to the Blits file converter plugin
- Fixed file extensions in the TypeScript boilerplate files
- Corrected the return value of the precompiler plugin


suresh-gangumalla (approved):
LGTM

PR #245, opened 2024-12-17
Fixes effects issues with direct obj assingment

suresh-gangumalla:
When complete object is assiged to state variable with equal operator then effects associated with in template on assining object properties are not triggering.

Here Object.assign is working as expected but users feels this is odd to do instead of assignment operator

Moved the object.assign operation to reactive proxy set trap so that it will invoke set traps of indiviual key


suresh-gangumalla:
Latest Changes are working as expected for object direct assignment.

PR #248, opened 2024-12-20
Dynamically set the width and height of a component

michielvandergeest:


suresh-gangumalla (approved):
LGTM

index.d.ts
```diff
@@ -244,11 +246,23 @@ declare module '@lightningjs/blits' {
...
+    $size: (dimensions: {
+      /**
+       * Component width
+       */
+      w: number,
+      /**
+       * Component height
+       */
+      h: number
+    }) => void
```
tuncayyildirtanuk, line 265: Hey @michielvandergeest,

Also If possible I would like to pass props to the element with only one object.

Just for example;

```
:props: $props, 
```

Example `props` objects
```
 const props1 = {
  fontFamily: 'Arial',
  fontSize: 20,
 }

const props2 = {
   x: 1,
   y: 2,
   w: 100,
   h: 100,
 }
```

michielvandergeest: Hey @tuncayyildirtanuk, technically this is possible (and we have an experimental branch for this available). But that solution has a performance impact, because changing a single prop value, leads to update attempts of other unchanged props as well.

We're not discarding the idea entirely, because I understand the use case from a developer perspective, but we have to be mindful of the performance impact.

Thanks for the suggestion in any case!

PR #249, opened 2024-12-23
Adding unknown for Index signature return type

suresh-gangumalla:
Input index singature Type is a function which returns void | undefined and all other custom properties Types should align with the index signature Type (including function return types as well). Here, the custom property intercept Type is a function that return value like KeyboardEvent, any, which are not part of index signature function return type so adding unknown as additional return type


michielvandergeest (approved):
LGTM

PR #252, opened 2025-01-09
Exclude index variable in forloop scope object

suresh-gangumalla:
If index variable is not declared in :for template then avoiding adding `index` variable to scope object of for-loop so that component may contain state or prop with same index variable which will be refferenced for all elements inside for-loop template.


michielvandergeest (approved):
LGTM

PR #258, opened 2025-01-16
Release 1.16.0

michielvandergeest:
- Fixed issue with component scope in Slots
- Fixed issue with sprites (regression since v1.14.0)
- Changed reactive state to only make plain objects reactive, and return custom Classes, Blits components, Renderer texture etc in raw form


erikhaandrikman (approved):
LGTM!

PR #260, opened 2025-01-17
Release 1.16.1

michielvandergeest:
- Fixed issue in certain browsers caused by route query params functionality (and optimized for performance)

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -156,16 +156,20 @@ export const navigate = async function () {
         holder.populate({})
         holder.set('w', '100%')
         holder.set('h', '100%')
+
+        const queryParamsData = {}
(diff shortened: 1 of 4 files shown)


src/router/router.js
```diff
@@ -156,16 +156,20 @@ export const navigate = async function () {
         holder.populate({})
         holder.set('w', '100%')
         holder.set('h', '100%')
+
+        const queryParamsData = {}
+        const queryParamsEntries = [...queryParams.entries()]
```
jfboeve, line 161: since queryParamsEntries is not used anywhere else... I don't think it necessary to do the [...entries] ?

jfboeve (approved):

PR #261, opened 2025-01-21
Bumped msdf-generator to 1.1.1

suresh-gangumalla:


michielvandergeest (changes requested):
Don't we also need an updated `package-lock.json` ?

PR #267, opened 2025-01-24
fix: register custom effects correctly

Daniel-Kusytsch:
Using custom effects fails because they are being registered as shaders; register them as effects instead.


michielvandergeest:
@Daniel-Kusytsch thanks for this PR.  But as you've seen a major overhaul of the shaders / effects system is upcoming. So we've kind of been waiting for that to land, to implement the custom shader registration properly.

PR #269, opened 2025-01-24
feat: add reactivity to router currentRoute object

vonvick:
#### What does this PR do?
This PR adds reactivity to the `currentRoute` property of the route object. Right now when the current route changes there is no way to know what it has changed to inside the app. This PR allows us to watch the `$router.currentRoute` object and react to the changes that occurs.

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -22,8 +22,10 @@ import { Log } from '../lib/log.js'
 import { stage } from '../launch.js'
 import Focus from '../focus.js'
 import Announcer from '../announcer/announcer.js'
+import { reactive } from '../lib/reactivity/reactive.js'
+import Settings from '../settings.js'
 
-export let currentRoute
+export const routeState = reactive({ currentRoute: null }, Settings.get('ReactivityMode'), true)
 export let navigating = false
 
 const cacheMap = new WeakMap()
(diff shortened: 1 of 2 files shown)


src/router/router.js
```diff
@@ -22,8 +22,10 @@ import { Log } from '../lib/log.js'
 import { stage } from '../launch.js'
 import Focus from '../focus.js'
 import Announcer from '../announcer/announcer.js'
+import { reactive } from '../lib/reactivity/reactive.js'
+import Settings from '../settings.js'
 
-export let currentRoute
```
michielvandergeest, line 26: I don't think we can just remove this export. There may be apps that use this value that would then break because of this change, and I don't think we should do a major bump for this feature alone.

Can we add some code to make this backwards compatible? Possibly with a deprecation log when using `currentRoute` directly (if possible).

michielvandergeest (changes requested):
This looks good, thanks @vonvick!

I think we should also make a change in the type definitions (i.e. deprecate currentRoute and type the new variable).

Also, it would be nice if we could add a line or 2 to the documentation about this (I'm aware currentRoute is currently also not documented, but this seems like a good opportunity to get this sorted)

michielvandergeest (approved):
LGTM. Thanks for contributing this :)

PR #273, opened 2025-01-29
Fixed issue with show when using also using alpha-attribute

jfboeve:


src/engines/L3/element.js
```diff
@@ -299,7 +299,7 @@ const propsTransformer = {
   },
   set show(v) {
     if (v) {
-      this.props['alpha'] = 1
+      this.props['alpha'] = !isNaN(this.raw['alpha']) ? this.raw['alpha'] : 1
```
michielvandergeest, line 302: can't this just be an undefined check for the raw alpha prop?

src/engines/L3/element.js
```diff
@@ -309,7 +309,9 @@ const propsTransformer = {
     }
   },
   set alpha(v) {
-    this.props['alpha'] = v
+    if(this.raw['show'] === true) {
```
michielvandergeest, line 312: what if the show-attribute isn't specified on an Element?

suresh-gangumalla (commented):
LGTM

PR #277, opened 2025-01-30
Added reactive router state to enable hooking up reactivity to route …

michielvandergeest:
…navigation changes.


suresh-gangumalla (approved):
LGTM

PR #279, opened 2025-01-31
Adds built-in component props data for the VSCode extension

uguraslan:
To provide more context-aware IntelliSense suggestions in the Blits VSCode extension, we need to maintain a data file containing a list of built-in component props and their details. Based on the data file created by @sandeep-vedam, I prepared a version that is better suited for the Blits VSCode extension.

The format:

```json
"clipping": {
  "attrType": "regular",
  "types": [
    "enum"
  ],
  "values": [
    "true",
    "false"
  ],
  "defaultValue": "false",
  "reactive": true,
  "description": "The clipping attribute cuts off the element positioning within specified height and width. Possible values are true or false.",
  "usedIn": [
    "Element",
    "Text",
    "Layout",
    "RouterView"
  ]
}
```

Each key in the first level defines the attribute/prop name followed by an info object containing:

Key | Description|
---|---
`attrType`    | Type of the attribute. Currently either `regular` or `event` (probably we will have more options for this one soon) |
`types`   | Possible types of attribute value (will be used for validation). For simple types, it is just string like `string`, `number`, `enum`. For complex types, an object definition must be used. (see `pivot` in the data file). Types are displayed in the info window when an attribute is selected from the suggestion list.|
`values`   | If the value type is `enum`, list of possible values.|
`defaultValue` | The default value for the attribute. This should be `null` if there is no default value.|
`reactive`   | Based on syntax alone, can the attribute also be suggested with a colon prefix (`:`). |
`description`   | The text displayed as the attribute description in the info window.|
`usedIn`   | In which built-in components, the attribute can be used.|

### DIFF
--- a/vscode/data/template-attributes.json
+++ b/vscode/data/template-attributes.json
@@ -1,0 +1,589 @@
+{
+  "x": {
+    "attrType": "regular",
+    "types": [
+      "number"
+    ],
+    "defaultValue": 0,
+    "reactive": true,
+    "description": "The x position of the Element in pixels, relative to its parent - allows negative values and decimals.",
+    "usedIn": [
+      "Element",
+      "Text",
+      "Layout",
+      "RouterView"
+    ]
+  },
+  "y": {
+    "attrType": "regular",
+    "types": [
+      "number"
+    ],
+    "defaultValue": 0,
+    "reactive": true,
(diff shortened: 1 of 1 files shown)


michielvandergeest (approved):
LGTM

PR #281, opened 2025-01-31
fix: avoid processing non-Blits template strings with precompiler

uguraslan:
Added content validation to check if template string starts with an HTML comment or a tag before processing it. This prevents issues with regular string values in objects that have a `template` property.


suresh-gangumalla (approved):
LGTM

michielvandergeest (approved):
LGTM!

PR #289, opened 2025-02-12
Disallows for loop in a root-level template tag

### LINKED ISSUE #288: Throw error when root level Element contains a for loop
Blits requires that a template has only one single root Element as a container. When a template does not have a single root a parsing error is thrown.

When a `:for` construct is used in a root level Element, then the parsing check passes, but it will generate unexpected results, because technically there isn't a single root anymore.

We should expand on the check to also consider the `:for`-attribute and throw an error indicating to wrap the for loop in container Element.

Maybe this can also be extended to the VScode extension showing an additional hint

uguraslan:
A template structure error is thrown when a for loop is used in a root-level template tag.

fixes #288


michielvandergeest (approved):
LGTM

PR #293, opened 2025-02-14
Fix/router history and state object update

vonvick:
#### What does this PR do?
This PR fixes the issue where the back navigation does not work as expected when the `currentRoute` is a dynamic route and the currentRoute object assigned to the `previousRoute` value is updated by reference when `currentRoute` changes. 

In addition to this fix, the route can now handle the back navigation even when the query parameters change.

The PR also adds more properties to the `$router.state` object, such as `queryParams, params, ` and updates the value referenced in the path property to be the correct hash path, even for dynamic routes.

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -177,16 +177,20 @@ export const navigate = async function () {
           queryParamsData[queryParamsEntries[i][0]] = queryParamsEntries[i][1]
         }
 
-        // merge props with potential route params, navigation data and route data to be injected into the component instance
-        const props = {
-          ...this[symbols.props],
-          ...route.params,
+        const routeData = {
           ...navigationData,
           ...route.data,
           ...queryParamsData,
         }
 
-        state.data = route.data
+        // merge props with potential route params, navigation data and route data to be injected into the component instance
+        const props = {
+          ...this[symbols.props],
+          ...route.params,
+          ...routeData,
+        }
+
+        state.data = routeData
 
         view = await route.component({ props }, holder, this)
         if (view[Symbol.toStringTag] === 'Module') {
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -343,7 +343,7 @@ const setOrAnimate = (node, transition, shouldAnimate = true) => {
 export const to = (location, data = {}, options = {}, queryParams = null) => {
   navigationData = data
   overrideOptions = options
-  let search
+  let search = ''
 
   if (queryParams) {
     search = new URLSearchParams(queryParams).toString()
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -26,11 +26,11 @@ import Announcer from '../announcer/announcer.js'
 
 export let currentRoute
 export const state = reactive({
-  path: null,
+  path: '',
   navigating: false,
-  queryParams: {},
+  data: {},
   params: {},
-  hash: null,
+  hash: '',
 })
 
 const cacheMap = new WeakMap()
@@ -44,8 +44,9 @@ let previousFocus
 export const getHash = () => {
   const hashParts = (document.location.hash || '/').replace(/^#/, '').split('?')
   return {
-    hash: hashParts[0],
+    path: hashParts[0],
     queryParams: new URLSearchParams(hashParts[1]),
+    hash: document.location.hash,
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/router/router.js
```diff
@@ -317,10 +337,15 @@ const setOrAnimate = (node, transition, shouldAnimate = true) => {
...
 }
 
-export const to = (location, data = {}, options = {}) => {
+export const to = (location, data = {}, options = {}, queryParams = {}) => {
   navigationData = data
   overrideOptions = options
-  window.location.hash = `#${location}`
+  let search = new URLSearchParams(queryParams).toString()
+
+  if (search) {
```
michielvandergeest, line 345: we can check specifically for an empty string versus a general truthy check

src/router/router.js
```diff
@@ -55,7 +61,20 @@ const isObject = (v) => typeof v === 'object' && v !== null
...
+  for (let i = 0; i < queryParamsEntries.length; i++) {
+    queryParamsData[queryParamsEntries[i][0]] = queryParamsEntries[i][1]
+  }
+
+  routeItem.queryParams = queryParamsData
+
+  return routeItem
+}
+
+export const matchHash = (path, queryParams = new URLSearchParams(), routes = []) => {
```
michielvandergeest, line 77: I would not instantiate always new URLSearchParams() here. That is relatively expensive operation, so I would only do that if we actually have queryParams

src/router/router.js
```diff
@@ -117,9 +137,10 @@ export const matchHash = (path, routes = []) => {
 export const navigate = async function () {
   state.navigating = true
   if (this.parent[symbols.routes]) {
-    const previousRoute = currentRoute
+    let previousRoute = currentRoute ? Object.assign({}, currentRoute) : undefined
```
michielvandergeest, line 140: Good catch :)

src/router/router.js
```diff
@@ -236,9 +251,14 @@ export const navigate = async function () {
...
         }
+
+        previousRoute = undefined
       }
 
-      state.path = route.path
+      state.rawPath = route.path
+      state.queryParams = route.queryParams
+      state.params = route.params
+      state.path = hash
```
michielvandergeest, line 261: something to think about: we could keep `path` and add `hash` (which will contain the populated in route path)

michielvandergeest (changes requested):
Overall looks good to me.

Left some minor comments.

src/router/router.js
```diff
@@ -128,7 +128,10 @@ export const matchHash = (path, queryParams = new URLSearchParams(), routes = []
       matchingRoute.data = {}
     }
 
-    matchingRoute = setRouteQueryParams(matchingRoute, queryParams)
+    if (queryParams) {
```
michielvandergeest, line 131: it's faster to do a specific `null` check

michielvandergeest (changes requested):

src/router/router.js
```diff
@@ -317,10 +337,15 @@ const setOrAnimate = (node, transition, shouldAnimate = true) => {
   })
 }
 
-export const to = (location, data = {}, options = {}) => {
+export const to = (location, data = {}, options = {}, queryParams = {}) => {
```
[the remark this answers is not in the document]
michielvandergeest: @vonvick regarding this: been thinking some more about this and it seems there's not really a need for passing data specifically as queryParams, when we can sprinkle it in as regular data

Then instead of watching queryPrams in the router state, we could probably just use data (and potentially watch that)

src/router/router.js
```diff
@@ -320,21 +337,18 @@ const setOrAnimate = (node, transition, shouldAnimate = true) => {
...
     // set indicator that we are navigating back (to prevent adding page to history stack)
     navigatingBack = true
-    let targetRoutePath = route.path
-    if (targetRoutePath.indexOf(':') > -1) {
-      Object.keys(route.params).forEach((item) => {
-        targetRoutePath = targetRoutePath.replace(`:${item}`, route.params[item])
-      })
-    }
-    to(targetRoutePath)
+    let targetRoutePath = route.hash
```
michielvandergeest, line 349: was it intentional to remove this?

michielvandergeest (approved):
LGTM! ⚡️

PR #295, opened 2025-02-18
Warning for non existing props/state variables used in template

suresh-gangumalla:


uguraslan:
Hey @suresh-gangumalla,

I've encountered some issues while testing the changes:

When running the Blits example app with `precompile` enabled, I'm getting this syntax error:
```bash
[plugin:preCompiler] missing ) after argument list
/blits-example-app-new/src/pages/KeyCodes.js
    at new Function (<anonymous>)
    at Object.default (file:///blits-suresh/src/lib/codegenerator/generator.js:63:13)
    at default (file:///blits-suresh/src/lib/precompiler/precompiler.js:54:34)
    at TransformContext.transform (file:///blits-suresh/vite/preCompiler.js:35:16)
    at Object.transform (file:///blits-example-app-new/node_modules/vite/dist/node/chunks/dep-b2890f90.js:44352:62)
    at async loadAndTransform (file:///blits-example-app-new/node_modules/vite/dist/node/chunks/dep-b2890f90.js:55026:29)
    at async viteTransformMiddleware (file:///blits-example-app-new/node_modules/vite/dist/node/chunks/dep-b2890f90.js:64434:32
```

With `precompile: false`, the app works but shows no errors about non-existing props, in the browser console.

I also created a new project using [the playground code](https://playground.lightningjs.io/#cly8CzZPqBw5)  - works fine with `precompile: true`.
However, when I copy `src/components/Button.js` from the example app to the new project, and use it, the same error reappears.

suresh-gangumalla:
@uguraslan  Yes, the problem lies in expression related to Math and logical. 
Implemented a logic to handle extracting only component state/prop variable from expression and then verify for their existence. 

Please verify agian with latest fix.

uguraslan:
Tested with `precompile=true` - the feature works as expected and catches undefined variables in templates that aren't defined as props or state variables.

When `precompile=false`, the checks don't work - I understand why it does not work, but not sure if this is expected behavior.

Also I don't know if this PR aims to cover all edge cases or just the common & basic scenarios for variable usage in templates.

If those limitations are intentional, then I think the changes are okay.

### Test Results:

_All variables in the tests do not exist as a prop or a state variable._

Template | Result | Status
--- | --- | ---
`<Circle color="$myColor">` | logs an error about `myColor` | ✅ 
`<Text :content="$myText" x="400" />` | logs an error about `myText` | ✅ 
`<Element :w="$width * 0.5" h="60" />` | no error log | ❌ 
`<Element :color="{top: $color2, bottom:'#000'}" w="300" h="200" />` | no error log | ❌ 
`<Element color="{top: $color2, bottom:'#000'}" w="300" h="200" />` | no error log | ❌
`<Text x="300" y="20" :content="'Page ' + $activePage + '/' + $totalPages" />` | no error log | ❌

suresh-gangumalla:
Hi @uguraslan  

Yes, You are true right now it works in precompile True mode. 

I have pushed new commit which handles most of the scenarios which I missed earlier. 

`$width`,
`$width * 0.5`,
`0.5 * $width`,
`100 - $width`,
`$width - 100`,
`100 * $width`,
`$width - 100`,
`100/$width`,
`$width/100`,
`100%$width`,
`$width%100`,
`$width || true`,
`true || $width`,
`!!$width`,
`2^100 + $width`,
`$width*@^100`,
`$width * $height`,
`$width * $height * 100`,
`100 * $width + $height`,
`:color='{top: $color2, bottom:'#000'}`,
`:color='{top: $color2, bottom:'$color4'}`,
`:color='{top: $color.top, bottom: $color.bottom * 1000}'`,
`100`,
`:effects=[$shader('radius', {radius: $radius})]`,
`{value: $x, duration: $duration, ease: $easing}`,
`{x: $mountX, y: $mountY}`,
`{top: $topColor, bottom: $bottomColor, left: $leftColor, right: $rightColor}`,

src/lib/codegenerator/generator.js
```diff
@@ -33,6 +34,28 @@ export default function (templateObject = { children: [] }) {
...
 
   counter = -1
+  isDev = this.isDev
+  if (isDev === true) {
+    ctx.renderCode.push(`
+      function propInComponent(prop, kind="dynamic") {
+        const property = prop.includes('.') ? prop.split('.')[0] : prop
+        if (kind === 'reactive' || prop.includes('.') === false) {
+          if (property in component === false) {
+            console.warn('Property ' +  property + ' was accessed during render but is not defined on instance')
```
michielvandergeest, line 44: Is there a way we can use the blits log functionality here, instead of console.log ?

src/lib/precompiler/precompiler.js
```diff
@@ -52,7 +51,7 @@ export default (source, filePath) => {
           const parsed = parser(templateContent, resourceName, null, filePath)
 
           // Generate the code
-          const code = generator.call({ components: {} }, parsed)
+          const code = generator.call({ components: {}, isDev: mode === 'development' }, parsed)
```
michielvandergeest, line 54: shouldn't we pass is dev mode as a function argument after `parsed` ?

michielvandergeest (changes requested):
we should probably also update the tests of the generator

the standard tests can remain the same if we default to dev mode off, and then it would be good to add 1 or 2 tests where the code to verify the variables is tested (i.e. running in dev mode)

src/lib/codegenerator/generator.js
```diff
@@ -54,6 +78,31 @@ export default function (templateObject = { children: [] }) {
...
+  const matches = value.match(regEx)
+  if (matches !== null) {
+    const shaderRegex = /\$shader/
+    return matches.filter((match) => !shaderRegex.test(match))
+  } else {
+    return false
+  }
+}
+
+const verifyVariables = function (value, renderCode, type = 'dymanic') {
```
jfboeve, line 93: type = 'dynamic' ?

suresh-gangumalla: Yes, it should be 'dynamic'. Thank you @jfboeve  
As I'm comparing type is 'reactive' or not, I haven't landed into any issue for dynamic case.
I will update it.

michielvandergeest (approved):
LGTM!

PR #312, opened 2025-03-14
Fixes reactivity guard comment formatting issue

uguraslan:
Added proper line breaks around injected code

Injected code now includes new lines before and after its content, preventing merging of comments with code blocks due to missing line separators.


michielvandergeest (approved):
LGTM!

PR #313, opened 2025-03-17
Persist $$ in plugins variables

suresh-gangumalla:
Previously the regular expression match return only single $ but for plugins variable expression should return double $$


michielvandergeest (approved):
LGTM

PR #315, opened 2025-03-17
Destoy View on Navigating Back with KeepAlive Set to True

suresh-gangumalla:
Fixes not removing router view, which configured as keep alive true, from cache map.


michielvandergeest (approved):
LGTM

PR #320, opened 2025-03-19
Updated refferencing cached route view on navigate

suresh-gangumalla:
1. On navigating using $route.to method, always create new instance of component eventhough its already cached by keepAlive
2. Not removing keepAlive pages from cache on back navigation
3. Destroying non keep alive pages on route change

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -179,7 +179,7 @@ export const navigate = async function () {
 
       let holder
       let routeData
-      let { view, focus } = cacheMap.get(route.hash) || {}
+      let { view, focus } = navigatingBack === true ? cacheMap.get(route.hash) || {} : {}
 
       if (!view) {
         // create a holder element for the new view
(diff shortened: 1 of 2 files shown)


src/router/router.js
```diff
@@ -177,9 +178,22 @@ export const navigate = async function () {
         route.transition = route.transition(previousRoute, route)
       }
 
+      if (navigatingBack === false && route.options && route.options.keepAlive === true) {
```
jfboeve, line 181: before this block I would recommend creating a variable:
```js
let viewManager = cacheMap.get(route.hash)`
```

The if statement you would do:
```js
if(navigatingBack === false && ....) {
    if(viewManager !== undefined) {
       viewManager.createEntry()
    } else {
       viewManager = new ViewManager()
       cacheMap.set(route.hash, viewManager)
    }
}
```

this allows us to not have to use cacheMap.get in the rest of the function.

src/router/router.js
```diff
@@ -177,9 +178,22 @@ export const navigate = async function () {
...
+        }
+      }
+
       let holder
       let routeData
-      let { view, focus } = cacheMap.get(route.hash) || {}
+      let { view, focus } =
+        navigatingBack === true && cacheMap.has(route.hash)
+          ? cacheMap.get(route.hash).getView() || {}
+          : {}
```
jfboeve, line 196: Now you can replace this with:
```js
let { view, focus } = navigatingBack === true && viewManager !== undefined && viewManager.getView() || {}
```

suresh-gangumalla:
Yes @michielvandergeest  this was handled in latest release so we are good to close this PR.

PR #327, opened 2025-03-25
Checking given value is non numeric or not

suresh-gangumalla:
1. Added regular expression that validate given value is valid numeric or not


suresh-gangumalla:
https://playground.lightningjs.io/#4oWqClD9tJSQ 

This playground verifies all possible numeric use cases

michielvandergeest (changes requested):
This looks good! Thanks

Can we add a test case for this use case to `generator.test.js`? Then we should be good to go

michielvandergeest (approved):
LGTM

PR #335, opened 2025-04-01
Fix/async request routing back

michielvandergeest:
Fix to address weird artefacts when routes are change during an async data request, as reported here: https://github.com/lightning-js/blits-example-app/issues/53


suresh-gangumalla (approved):
LGTM

PR #336, opened 2025-04-01
Release v1.25.1

michielvandergeest:
- Fixed issue with page rendering after being destroyed when data is loaded by an async request


suresh-gangumalla (approved):
LGTM

PR #341, opened 2025-04-09
Addressed renderer package import issue in blits

suresh-gangumalla:


src/launch.js
```diff
@@ -19,17 +19,30 @@ import Settings from './settings.js'
...
+  let rendererPackageInfo
+
+  try {
+    // Dynamically import the renderer package.json
+    rendererPackageInfo = await import('../../renderer/package.json')
+    if (rendererPackageInfo !== undefined) {
+      Log.info('Renderer Version ', rendererPackageInfo.version)
+    }
+  } catch (e) {
+    // Ignore
```
michielvandergeest, line 36: In case of error we could use the renderer version from the blits package.json?

michielvandergeest (changes requested):
looks good, thanks!

a few minor comments

src/launch.js
suresh-gangumalla: Yeah sounds good, l will include that.

michielvandergeest (approved):

PR #343, opened 2025-04-10
Added stop function on speech utterance to allow developers to stop a…

michielvandergeest:
… specific message.

### DIFF
--- a/src/announcer/announcer.js
+++ b/src/announcer/announcer.js
@@ -21,6 +21,8 @@ let active = false
 let count = 0
 const queue = []
 let isProcessing = false
+let currentId = null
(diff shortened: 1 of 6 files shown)


src/announcer/announcer.js
```diff
@@ -84,27 +95,37 @@ const processQueue = async () => {
...
+  if (delay) {
     setTimeout(() => {
       isProcessing = false
+      currentId = null
       resolveFn('finished')
       processQueue()
     }, delay)
-    return
+  } else {
+    if (debounce) clearTimeout(debounce)
```
jfboeve, line 110: debounce !== null

PR #350, opened 2025-04-11
Feat/storage plugin

sandeep-vedam:
Added Storage Plugin


michielvandergeest (changes requested):
I think the storage plugin should be completely optional and follow the same implementation style as the theme and language plugins

docs/plugins/storage.md
```diff
@@ -0,0 +1,76 @@
...
+
+### set
+
+Saves a key-value combination in storage.
+
+```js
+this.$storage.set(key, value)
+```
+
+The key is expected to be a `String`. The value can be a `String`, `Object`, `Boolean` or `Array`.
```
jfboeve, line 48: No numbers?

michielvandergeest: @sandeep-vedam I think @jfboeve was referring to the _value_ also being a number and not so much the key?

jfboeve:
/src/plugin/Storage directory should be changed to /src/plugin/storage to match the rest of the repository

src/plugins/Storage/localCookie.mjs
```diff
@@ -0,0 +1 @@
...eys:this._getCookieKeys}}_checkIfLocalStorageWorks(){if("undefined"==typeof localStorage)return!1;try{return localStorage.setItem("feature_test","yes"),"yes"===localStorage.getItem("feature_test")&&(localStorage.removeItem("feature_test"),!0)}catch(e){return!1}}_getItemLocalStorage(e){return window.localStorage.getItem(e)}_setItemLocalStorage(e,t){return window.localStorage.setItem(e,t)}_removeItemLocalStorage(e){return window.localStorage.removeItem(e)}_clearLocalStorage(){return window.localStorage.clear()}_getLocalStorageKeys(){return Object.keys(window.localStorage)}_getItemCookie(e){var t=document.cookie.match(RegExp("(?:^|;\\s*)"+function(e){return e.replace(/([.*+?\^${}()|\[\]\/\\])/g,"\\$1")}(e)+"=([^;]*)"));return t&&""===t[1]&&(t[1]=null),t?t[1]:null}_setItemCookie(e,t){var o=new Date,r=new Date(o.getTime()+15768e7);document.cookie=`${e}=${t}; expires=${r.toUTCString()};`}_removeItemCookie(e){document.cookie=`${e}=;Max-Age=-99999999;`}_clearCookies(){document.cookie.split(";").forEach(e=>{document.cookie=e.replace(/^ +/,"").replace(/=.*/,"=;expires=Max-Age=-99999999")})}_getCookieKeys(){return document.cookie.split(";").map(e=>e.split("=")[0])}}export default localCookie;
```
michielvandergeest, line 1: We don't need this file to be minified, it can just be the src file from the localCookie dependency (with changes where needed).

Also we don't need the mjs file extension, just .js is fine.

michielvandergeest (changes requested):

jfboeve (approved):

michielvandergeest (approved):
LGTM!

PR #354, opened 2025-04-15
Fixed issue where global watchers weren't properly cleared and were k…

michielvandergeest:
…ept in memory.


suresh-gangumalla (approved):
LGTM

Issue #349, opened 2025-04-11
bug: nested layout: the outer layout doesn't always update its dimensions after the inner does


GauthierBosson:
Hello,

I have a component which uses nested layout and the outer layout doesn't always trigger an update when the nested layout recalculates its dimensions, resulting in bad placements of the elements.

here is the component code:

```javascript
import Blits from '@lightningjs/blits'
import Thumbnail from '../atoms/Thumbnail'
import Tag from '../atoms/Tag'

export default Blits.Component('NewTeaser', {
  components: {
    Thumbnail,
    Tag
  },
  template: `
    <Element x="0" y="0">
      <Thumbnail
        ref="thumbnail"
        :w="$$theme.get('thumbnail.sizes.16:9.width')"
        :h="$$theme.get('thumbnail.sizes.16:9.height')"
      />
      <Layout :x="$detailsPositionX" direction="vertical" gap="14" @updated="$onUpdated">
        <Tag text="NEW" />
        <Layout direction="vertical" gap="4" @updated="$onUpdated">
          <Text
            :font="$$theme.get('fonts.headline.medium.font')"
            :size="$$theme.get('fonts.headline.medium.size')"
            :lineheight="$$theme.get('fonts.headline.medium.lineHeight')"
            :letterspacing="$$theme.get('fonts.headline.medium.letterSpacing')"
            :content="$item.title"
            :maxwidth="$MAX_WIDTH"
          />
          <Text
            :font="$$theme.get('fonts.body.large.font')"
            :size="$$theme.get('fonts.body.large.size')"
            :lineheight="$$theme.get('fonts.body.large.lineHeight')"
            :letterspacing="$$theme.get('fonts.body.large.letterSpacing')"
            :content="$item.description"
            :maxwidth="$MAX_WIDTH"
          />
        </Layout>
        <Text
          :size="$$theme.get('fonts.body.large.size')"
          :font="$$theme.get('fonts.body.large.font')"
          :lineheight="$$theme.get('fonts.body.large.lineHeight')"
          :letterspacing="$$theme.get('fonts.body.large.letterSpacing')"
          :content="$item.oneLiner"
        />
      </Layout>
    </Element>
  `,
  props: ['item'],
  state() {
    return {
      MAX_WIDTH: 1000,
    }
  },
  computed: {
    detailsPositionX(): number {
      return this.$theme.get('thumbnail.sizes.16:9.width') + 32
    }
  },
  hooks: {
    focus() {
[...]

PR #355, opened 2025-04-15
Added trigger to update parent layout in the case of nested layout.

michielvandergeest:
Fixes https://github.com/lightning-js/blits/issues/349


suresh-gangumalla (approved):
LGTM

PR #358, opened 2025-04-18
Renamed beforeAll router hook to beforeEach

michielvandergeest:
The router hook function ended up being called beforeAll in the code, whereas I think we intended to name it beforeEach


jfboeve (approved):
LGTM!

PR #362, opened 2025-04-24
Feat/inline text reactivity

suresh-gangumalla:
1. Added support for inline text reactivity
2. Added logic to parse text interpolation `({{ }})`
3. Added new test cases to parser
4. Updated code-generator failed test cases

Created a sample playground example which points to current PR branch and demonstrate few possible inline text combinations with JS expressions from state object, Math and logical. 

https://playground.lightningjs.io/#sgqS4HeLwk9A

please add missing scenarios to capture failures at this stage :)

### DIFF
--- a/src/lib/codegenerator/generator.js
+++ b/src/lib/codegenerator/generator.js
@@ -187,6 +187,31 @@ const generateElementCode = function (
     }
   })
 
+  if (templateObject[Symbol.for('tagContent')] !== undefined) {
+    const val = templateObject[Symbol.for('tagContent')]
+
+    // Check if the string contains interpolation (`{{ }}`).
+    const regex = /\{\{\s*.+?\s*\}\}/
+    const containsInterpolation = regex.test(val)
+
+    if (containsInterpolation === false) {
+      // Inline text is treated as static content when interpolation is not defined.
+      renderCode.push(`elementConfig${counter}['content'] = '${val}'`)
+    } else {
+      const output = parseTagContent(val, options.component)
+      renderCode.push(`elementConfig${counter}['content'] = ${output}`)
+
+      const variableRegEx = /\{\{\s*[^}]*\$\S*\s*\}\}/
+      // Check if the interpolation contains $ variable
+      const isReactive = variableRegEx.test(val)
+      if (isReactive === true) {
+        this.effectsCode.push(`
+          ${elm}.set('content', ${output})
+        `)
+      }
+    }
+  }
+
   if (options.holder === true) {
     renderCode.push(`
     const skip${counter} = []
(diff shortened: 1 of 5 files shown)


src/lib/codegenerator/generator.js
```diff
@@ -748,6 +773,57 @@ const cast = (val = '', key = false, component = 'component.') => {
...
+  let escapedVal = val
+    .replace(/\\\\/g, '__DOUBLE_BACKSLASH__')
+    .replace(/(^|[^\\])'/g, "$1\\'")
+    .replace(/__DOUBLE_BACKSLASH__/g, '\\\\')
+
+  const dynamicParts = /\{\{\s*.+?\s*\}\}/g
+  const matches = [...escapedVal.matchAll(dynamicParts)]
+
+  if (matches.length) {
+    const isValStartsWithBrace = /^\{\{/.test(escapedVal)
```
jfboeve, line 787: replace with escapedVal.startsWith('{{'), this is faster that using the regex test.

src/lib/codegenerator/generator.js
```diff
@@ -748,6 +773,57 @@ const cast = (val = '', key = false, component = 'component.') => {
...
+    const isValEndsWithBrace = /\}\}$/.test(escapedVal)
```
jfboeve, line 788: replace with escapedVal.endsWith('}}'). same as with starts with.

src/lib/codegenerator/generator.js
```diff
@@ -748,6 +773,57 @@ const cast = (val = '', key = false, component = 'component.') => {
...
+  // unescaped single quotes must be escaped while preserving escaped backslashes
+  let escapedVal = val
+    .replace(/\\\\/g, '__DOUBLE_BACKSLASH__')
+    .replace(/(^|[^\\])'/g, "$1\\'")
+    .replace(/__DOUBLE_BACKSLASH__/g, '\\\\')
+
+  const dynamicParts = /\{\{\s*.+?\s*\}\}/g
+  const matches = [...escapedVal.matchAll(dynamicParts)]
+
+  if (matches.length) {
```
jfboeve, line 786: matches.length > 0

src/lib/codegenerator/generator.js
```diff
@@ -748,6 +773,57 @@ const cast = (val = '', key = false, component = 'component.') => {
...
+    for (let matchObj of matches) {
+      const { 0: match, index } = matchObj
+      const isMatchAtStart = index === 0
+      const isMatchAtLast = val[index + match.length] === undefined ? true : false
+
+      let parsedMatch = match
+
+      const replaceDollar = /\$(\$(?=\$)|\$?)/g
+      const dollarMatches = [...parsedMatch.matchAll(replaceDollar)]
+      if (dollarMatches.length) {
```
jfboeve, line 798: dollarMatches.length > 0

src/lib/codegenerator/generator.js
suresh-gangumalla: Thank you @jfboeve, Jsbench results that `.startsWith()` is 56% faster than regular expression test.

jfboeve: Yw @suresh-gangumalla, the reason why it's faster is because regex does not have to be instantiated first :)

src/lib/codegenerator/generator.test.js
```diff
@@ -2430,8 +2491,8 @@ test('Generate code for a template with inline dynamic Text embedded in static t
     'Generator should return a render function with the correct code'
   )
   assert.ok(
-    Array.isArray(actual.effects) && actual.effects.length === 0,
-    'Generator should return an empty effects array'
+    Array.isArray(actual.effects) && actual.effects.length === 1,
+    'Generator should return effects array with length 1'
```
michielvandergeest, line 3135: @suresh-gangumalla Maybe we should also test the body of the effect function ? And not just whether an effect was generated?

PR #364, opened 2025-04-29
Feature/v2 props object

michielvandergeest:


src/component/setup/props.js
```diff
@@ -16,44 +16,55 @@
...
+
+  if (Array.isArray(props) === true) {
+    Log.warn(
+      'Defining props as an Array has been deprecated and will stop working in future versions. Please use the new notation instead (an object with key values pairs).'
+    )
+    const propLength = props.length
+    for (let i = 0; i < propLength; i++) {
+      const prop = props[i]
+      if (typeof prop === 'string') {
+        out[prop] = { default: undefined }
```
jfboeve, line 32: I think these don't need the be storing objects with when you're only using 1 value in them.. Or am I missing something?

michielvandergeest: this is for backwards compatibility - the old syntax could just be an array, in the new setup we're expecting an object. This piece of code rewrites it to the expected format

suresh-gangumalla (approved):

PR #367, opened 2025-05-15
Added error funtion to Router configuration

suresh-gangumalla:
Added error function to Router configuration which can return string | void, returning string results to navigating to given new route path


michielvandergeest:
Closing this PR as it's already added in another PR

PR #368, opened 2025-05-19
feat: add documentation for range variable in for-loop

WoodingMP:
I found this range variable in one of the release notes but couldn't find documentation for how it works anywhere so just wanted to share what I found using it to hopefully help others out. If there's more functionality here that I've missed please add/change what I've written up


wouterlucas (approved):
Nice! Was just looking for this 😄

michielvandergeest (approved):
Looks great! Thanks for adding this 😊

PR #373, opened 2025-05-22
Add JSDoc Type definitions

wouterlucas:
Work in progress - started as a way to understand Blits on a deeper level. Now kind of snowballing into a thing.

Adding JSDoc style types for autocompletion and typesafety.

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,15 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "ESNext",
+    "allowJs": true,
+    "checkJs": true,
+    "moduleResolution": "node",
+    "baseUrl": "./",
+    "paths": {
+      "@/*": ["src/*"]
+    }
+  },
+  "include": ["src"],
+  "exclude": ["node_modules", "dist"]
+}
--- a/src/component.js
+++ b/src/component.js
@@ -44,11 +44,52 @@ const required = (name) => {
 }
 
 /**
+ * @typedef {function} BlitsComponentFactory
+ * @param {object} opts - The options for the component instance
+ * @param {BlitsComponent} parentEl - The parent element for the component instance
+ * @param {BlitsComponent} parentComponent - The parent component for the component instance
+ * @param {BlitsComponent} rootComponent - The root component for the component instance
+ * @returns {BlitsComponent} - The component instance
+ */
+
+/**
+ * @typedef {import('@/engines/L3/element.js').BlitsElement} BlitsElement
+ */
+
+/**
+ * @typedef {object} BlitsComponent
+ * @property {string} componentId - The unique identifier for the component instance
+ * @property {object} lifecycle - The lifecycle object for the component instance
+ * @property {object} parent - The parent component of the current component instance
+ * @property {object} rootParent - The root parent component of the current component instance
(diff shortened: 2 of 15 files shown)


wouterlucas:
Since I'm not modifying anything code wise, I hope we can move this along and incrementally improve it instead of doing it a one shot thing that includes everything.

Would appreciate your review 👀 Michiel

jsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "ESNext",
+    "allowJs": true,
+    "checkJs": true,
+    "moduleResolution": "node",
+    "baseUrl": "./",
+    "paths": {
+      "@/*": ["src/*"]
+    }
```
michielvandergeest, line 11: `@/`-alias imports are not used within the Blits code base

jsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "include": ["src"],
+  "exclude": ["node_modules", "dist"]
```
michielvandergeest, line 14: `dist` can be removed, the blits repo itself doesn't have a dist folder. We may want to exclude the `packages` folder (for now), unless it doesn't cause any issues

src/component.js
```diff
@@ -43,12 +43,150 @@ const required = (name) => {
...
+ * @property {number} nodeId - ID of the CoreNode, if available.
+ * @property {string|null} ref - Ref name (if defined).
+ * @property {function(Object):void} populate - Initializes the element with props and hooks.
+ * @property {function(string, any):void} set - Updates a single property.
+ * @property {function(string, any, Object):void} animate - Animates a property with transition options.
+ * @property {function():void} destroy - Destroys the underlying node and cancels transitions.
+ * @property {function(any): any} triggerLayout - Triggers a layout update for the element.
+ *
+ * @typedef {Object} BlitsLifecycle
+ * @property {BlitsComponent} component - The component WVB do we need this here again?
```
michielvandergeest, line 107: yes it's the scope and this makes sure lifecycle scope is unique per component instance

src/router/router.js
```diff
@@ -129,14 +185,34 @@ export const matchHash = (path, routes = []) => {
...
-    let route = matchHash(path, this.parent[symbols.routes])
+    const routeOrFalse = matchHash(path, this.parent[symbols.routes])
+
+    if (!routeOrFalse) {
+      // If no route is found, probably handle with a generic route?
+      return
+    }
+
+    /** @typedef {Route} */
+    let route = routeOrFalse
```
michielvandergeest, line 215: can we remove this code change from this PR? There is an issue open regarding error handling when route doesn't exist, and a feature underway to handle this case gracefully

michielvandergeest (changes requested):
This is great! Thanks for making this first (huge) start! We can incrementally build upon this.

Few minor comments and change requests.

jsconfig.json
wouterlucas: ah right - this my editor being smart. Will remove! thanks

jsconfig.json
wouterlucas: Yep - will update! thanks!

src/component.js
wouterlucas: OK - will remove comment 😄

PR #375, opened 2025-05-28
fix: set isProcessing to false when the utterance is cancelled

vonvick:
# What does this PR do?

This PR sets `isProcessing` to false when the utterance is cancelled, so that the queue is free and can process the next utterance


michielvandergeest (approved):
LGTM!

PR #388, opened 2025-06-23
Decouple msdf-generator dependency

uguraslan:
Removes `@lightningjs/msdf-generator` from direct dependencies.

The Vite plugin now checks for this package at runtime. If font generation is required and the package is missing, it prints installation instructions and exits to prevent a broken build. 

Adds the package as a dev dependency in boilerplate codes.


suresh-gangumalla (approved):
LGTM

PR #411, opened 2025-07-09
Fix: announcer utterance onend

vonvick:


michielvandergeest (approved):
LGTM! Thanks for figuring this out :)

PR #414, opened 2025-07-09
Fix/docs transition - correct example and description of args for transition progress method

che-effe:
I noticed that the transition config object's progress key handler method actually takes in a different set of arguments than what is documented.
Updating this now, and I believe someone on my team may also suggest a change to have an object be passed as the argument to make it easier to deconstruct within the receiving method.

args observed in the handler method

- component, 
- property, 
- currentProgress, 
- prevProgress,

Playground demo of this
https://playground.lightningjs.io/#ErYUaCZjAGY4

### DIFF
--- a/docs/transitions_animations/transitions.md
+++ b/docs/transitions_animations/transitions.md
@@ -130,7 +130,7 @@ export default Blits.Component('Gold', {
 
 It is also possible to keep track of the entire progress of a transition. Every frametick during a transition (ideally once every 16ms), the renderer reports the progress of the transition. You can hook into this event by specifying a `progress` key on the transition configuration object, with a function to excute.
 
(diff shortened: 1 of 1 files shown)


docs/transitions_animations/transitions.md
```diff
@@ -141,7 +141,7 @@ export default Blits.Component('Gold', {
   `,
   ///
   methods: {
-    transitionProgress(progress, previousProgress) {
+    transitionProgress(component, prop, progress, previousProgress) {
```
michielvandergeest: I don't think the first argument is the `component`, but I think it's the element that the transition is being applied on? You'r totally right about the second argument being the `property` being affected though!

michielvandergeest (approved):
LGTM!

PR #415, opened 2025-07-09
Refactor progress params

ImCoolNowRight:
This PR consolidates the params being used for progress updates in transitions into a single object to make it easier for developers to only retrieve the ones they need.

### DIFF
--- a/docs/transitions_animations/transitions.md
+++ b/docs/transitions_animations/transitions.md
@@ -141,7 +141,7 @@ export default Blits.Component('Gold', {
   `,
   ///
   methods: {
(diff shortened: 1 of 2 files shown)


docs/transitions_animations/transitions.md
```diff
@@ -141,7 +141,7 @@ export default Blits.Component('Gold', {
   `,
   ///
   methods: {
-    transitionProgress(element, prop, progress, previousProgress) {
+    transitionProgress({element, prop, progress, previousProgress}) {
```
che-effe, line 144: It might also be a good idea to update the description of the args up on line 133 in this same file.

PR #417, opened 2025-07-10
Refactor progress params

ImCoolNowRight:
This PR consolidates the params being used for progress updates in transitions into a single object to make it easier for developers to only retrieve the ones they need.


michielvandergeest (approved):
LGTM

PR #421, opened 2025-07-15
Test-Coverage-Tuesday- Added Additional Tests For Element.js

suresh-gangumalla:
Improved the code-coverage from 58 to 75.*


src/engines/L3/element.test.js
```diff
@@ -568,12 +568,299 @@ test('Element - Set `show` property as false', (assert) => {
...
+  const el = createElement()
+
+  el.set('w', { transition: { value: 100 } })
+
+  assert.equal(el.node['width'], 100, 'Node width parameter should be set')
+  assert.equal(el.props.props['width'], 100, 'Props width parameter should be set')
+  assert.end()
+})
+
+test('Element - Listen to transition start callback on `w` prop changes', (assert) => {
```
michielvandergeest, line 582: In this test, where are we checking if the start callback is fired? Shouldn't we put a Spy on that callback and verify that it's called?

src/engines/L3/element.test.js
```diff
@@ -568,12 +568,299 @@ test('Element - Set `show` property as false', (assert) => {
...
+
+  el.set('w', { transition: { value: 50 } })
+  el.set('w', { transition: { value: 100 } })
+
+  assert.equal(el.node['width'], 100, 'Node width parameter should be set')
+  assert.equal(el.props.props['width'], 100, 'Props width parameter should be set')
+  assert.end()
+})
+
+test('Element - Trigger layout on parent element through transition', (assert) => {
```
michielvandergeest: How are we testing the invocation of the layout function when the `__layout: true` property has been set?

src/engines/L3/element.test.js
```diff
@@ -568,12 +568,299 @@ test('Element - Set `show` property as false', (assert) => {
...
+  assert.equal(el.props.props['clipping'], true, 'Props clipping parameter should be set')
+
+  el.set('overflow', true)
+  assert.equal(el.node['clipping'], false, 'overflow attribute should set node clipping parameter')
+  assert.equal(el.props.props['clipping'], false, 'Props clipping parameter should be set')
+
+  assert.end()
+})
+
+test('Element - Transition an element property with progress callback', (assert) => {
```
michielvandergeest, line 882: Here we should also put a Spy on the progress callback and ensure that it's called multiple times?

src/engines/L3/element.test.js
```diff
@@ -568,12 +568,299 @@ test('Element - Set `show` property as false', (assert) => {
...
+  const el = createElement({ props: { __layout: true } })
+
+  el.set('w', { transition: { value: 100, progress: () => {} } })
+
+  assert.equal(el.node['width'], 100, 'Node width parameter should be set')
+  assert.equal(el.props.props['width'], 100, 'Props width parameter should be set')
+  assert.end()
+})
+
+test('Element - Transition an element property with end callback', (assert) => {
```
michielvandergeest, line 896: Same as above about the start and progress callbacks

src/engines/L3/element.test.js
```diff
@@ -568,12 +568,299 @@ test('Element - Set `show` property as false', (assert) => {
...
+  el.destroy()
+  assert.equal(el.node, null, 'Node should set to null')
+  assert.equal(el.component, undefined, 'Component should be deleted from element')
+  assert.equal(el.props, undefined, 'Props should be deleted from element')
+  assert.equal(el.config, undefined, 'Config should be deleted from element')
+  assert.equal(el.triggerLayout, undefined, 'TriggerLayout should be deleted from element')
+  assert.end()
+})
+
+test('Element - Element with error callback', (assert) => {
```
michielvandergeest, line 922: Here I'm also missing the validation that the error function is actually being called

src/engines/L3/element.test.js
```diff
@@ -725,23 +892,28 @@ test('Element - Create Text Node with supported props 3', (assert) => {
...
-  el.set('w', { transition: { value: 100, progress: () => {} } })
+  const el = createElement()
+
+  const progSpy = sinon.spy()
+
+  el.set('w', { transition: { value: 100, progress: progSpy } })
 
-  assert.equal(el.node['width'], 100, 'Node width parameter should be set')
   assert.equal(el.props.props['width'], 100, 'Props width parameter should be set')
+  // assert.equal(progSpy.callCount, 10, 'Transition progress callback should be called 10 times')
```
michielvandergeest, line 892: I don't think that we can guarantee an exact amount of progress callbacks, it will depend on the framerate. But it is expected to be called multiple times with a higher progress value (something between 0 and 1). Maybe we can test for that and get enough confidence from that

suresh-gangumalla: Agreed if real renderer instance animation library is emiting this progress event then we can't gurantee, but here we are creating custom animator and making sure it will emit 'tick' event multiple times, I have custom logic in CustAnimator class.

suresh-gangumalla: Forgot about call count, I just found atleast one time also spy callback is not triggering in Progress and end use case, It is working as expected in start callback.

PR #422, opened 2025-07-21
Fixed issue with router throwing an error when a route doesn't exist …

michielvandergeest:
…instead of handling it gracefully.


vonvick (approved):

suresh-gangumalla (approved):
LGTM

PR #423, opened 2025-07-21
Refactor/focus

michielvandergeest:
Refactor for focus to make it work more reliably"

- Fixes issue with not all pages in the chain calling unfocus hook when programatically changing the focus (in an emit for example).
- Ensures that $hasFocus is always set correctly.
- Optimizes focus chains

### DIFF
--- a/src/focus.js
+++ b/src/focus.js
@@ -26,6 +26,20 @@ let setFocusTimeout
 
 export const keyUpCallbacks = new Map()
 
+/**
+ * Safely unfocus a component with error handling
+ * @param {Object} component - The component to unfocus
+ */
+const safeUnfocus = (component) => {
+  try {
+    if (component && typeof component.unfocus === 'function') {
+      component.unfocus()
+    }
+  } catch (error) {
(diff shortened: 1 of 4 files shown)


src/focus.js
```diff
@@ -38,56 +39,106 @@ export default {
...
+const setFocus = (component, event) => {
+  Log.info(
+    '\nFocus chain:\n',
+    focusChain.map((c, index) => '\t'.repeat(index) + '↳ ' + c.componentId).join('\n')
+  )
+  focusedComponent = component
+  component.lifecycle.state = 'focus'
+
+  if (event instanceof KeyboardEvent) {
+    const internalEvent = new KeyboardEvent('keydown', event)
```
wouterlucas, line 151: interesting 😅 was it always a KeyboardEvent? Or are we piggybacking?

wouterlucas (approved):
Nice!

src/focus.js
michielvandergeest: Yep - we're (optionally) passing the actual keyboard event when passing on focus. So this has always been in there like this 😄

PR #428, opened 2025-07-24
Feature/delegate focus on destroy

michielvandergeest:
Guards for a potential loss of focus in the use case where a component that currently has focus (for example a tile in a list) is destroyed due to a data change.

Now it will gracefully delegate the focus to the parent


suresh-gangumalla (approved):
LGTM

PR #433, opened 2025-07-30
feat: Optionally skip passing focus to RouterView on navigate

WoodingMP:
In our application we've got a navigation bar that sits about the RouterView in our App component, the idea being that this navigation bar is available on every page, and has the logic to navigate between them while remaining visible the whole time.

We have a pain point with the current router implementation when navigating between pages in that anytime you use `$router.to`, focus automatically gets passed to the RouterView and then along to the page component. For us this means that our nav gets unfocused every time we swap pages rather than what we need which is for the nav bar to remain in focus until you scroll down into a page.

To support this use case, I've added a check for a `passFocus` option when navigating. This won't affect the current logic when unused, but if set to `false` as part of the router options, the current focus will not change when navigating.

Example:
`this.$router.to('home', {id: 123456}, {passFocus: false});`

This would prevent us needing to constantly fight the current focus and try and refocus the nav when going between pages

### DIFF
--- a/docs/router/basics.md
+++ b/docs/router/basics.md
@@ -96,6 +96,8 @@ export default Blits.Component('Poster', {
 
 Whenever you navigate to a new page, the URL hash will automatically be updated. Unless specified otherwise, navigating to a new page, will add that route to the history stack. The `back` input action is automatically wired up to navigate back down the history stack.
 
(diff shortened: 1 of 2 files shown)


WoodingMP:
Thanks @michielvandergeest! I leave the naming in your capable hands it was just the first thing that popped to mind haha

michielvandergeest (commented):
LGTM!

I think the name you came up with, is perfectly in line with the existing `inHistory` and `keepAlive`, so `passFocus` feels like a goon option to me.

Thanks again for adding this functionality 🙂

PR #434, opened 2025-07-31
Fix/router navigate data

michielvandergeest:
Fixes issue where data passed as part of router navigation is not being updated in the props (reported via Discord)


suresh-gangumalla (approved):
LGTM

PR #447, opened 2025-08-12
Upgraded below dev dependencies versions

suresh-gangumalla:
1. updated c8 package from 8.0.1 to 9.1.0
 2. udpated eslint package from 8.57.1 to 9.33.0


eslint.config.cjs
```diff
@@ -0,0 +1,72 @@
+const { defineConfig, globalIgnores } = require('eslint/config')
```
michielvandergeest, line 18: we need to add license headers to this file (as we did in the `.eslintrc.cjs`).

this new file name is the naming convention used in the new version of ESlint we're using now?

suresh-gangumalla: I will update license header and Yes, I have used utility package provided by ESlint to generate config.js from rc file as part of migration from 8 to 9.

PR #462, opened 2025-08-25
Added reuse component feature to router.

michielvandergeest:
With this feature, when navigating to a route which uses the same page component we can now reuse the same component instead of creating a new instance and destroying the old one.


suresh-gangumalla (approved):
LGTM

Issue #461, opened 2025-08-22
Router option overrides don't get cleared on multiple navigations


WoodingMP:
Hi I'm trying to the router to navigate between pages with option overrides on top of the configuration set in the initial router object.

The issue I'm find is that if I set an override like 
`this.$router.to('page', {}, {inHistory: false})`
the override will remain every time I navigate to that page unless I specifically reset it's value with
`this.$router.to('page', {}, {inHistory: true})`

I'm not sure that this is intended, as it feels like it'd make more sense for overrides to only apply to the single instance of to or navigate that they're used in, without overwriting the initial configuration.

To help demonstrate I've setup a playground:
https://playground.lightningjs.io/#RVMs5ESST0HZ

Try going through the steps outlined in the input section, following to 1-6 keys with the back presses in-between. You'll see that when I navigate without any options, history works as intended. Then after I set `inHistory:false` the movies page is no longer part of the history, even when I run `this.$router.to('movies')` without an overrides

michielvandergeest:
This should be fixed in 1.39.0 release of blits. Closing for now, feel free to reopen or create a new issue if the problems persists

PR #466, opened 2025-08-28
Clone route objects to avoid passing stale props

WoodingMP:
This is related to an issue I posted https://github.com/lightning-js/blits/issues/461 where router options were getting stored in the original route object on every navigation - causing these assumed temporary options to be used every time you try navigate to a page. This example shows the problem using inHistory
https://playground.lightningjs.io/#RVMs5ESST0HZ

There was another related issue I started noticing when trying to use keepAlive with dynamic routes, where the pages were being saved in cache using the wrong hash. This is also because of a stale route object being used in `removeView` when setting the cache. I've setup another example here to demonstrate the problem - use 1/2/3 to nav between pages and see that sometimes the wrong number appears on screen because the cache has the wrong view attached
https://playground.lightningjs.io/#avFQk5g1Wwlr

To fix both these problems, instead of passing routes by reference, we can clone the starting values so that the correct options/views are being used during the navigate function. One of these fixes looks like it had been used before but was commented out, so if there's a particular reason this won't work let me know! It seemed to fix my issue but might be causing some other problems.

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -176,6 +176,7 @@ export const matchHash = (path, routes = []) => {
   }
 
   if (matchingRoute) {
+    matchingRoute = Object.assign({}, matchingRoute) // clone to avoid overwriting original options
     matchingRoute.options = { ...defaultOptions, ...matchingRoute.options, ...overrideOptions }
     if (!matchingRoute.data) {
       matchingRoute.data = {}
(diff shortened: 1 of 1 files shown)


michielvandergeest:
hey @WoodingMP - thanks for this contribution. As we've discussed offline already, the Object.assign was commented out because it seemed to be responsible for memory leaks. Instead we have refactored the router code to prevent the issue you've mentioned. This has been released in Blits v1.39.0, so closing this PR.

Thanks again for the contribution. Much appreciated!

PR #468, opened 2025-08-28
Test coverage tuesday/router state tests

il-sairamg:
This PR introduces a unit test to verify that the router state (state.path, state.params, state.data) is correctly set and accessible inside the init() lifecycle hook of a component.

Changes:

Updated navigate() to set state.path, state.params, state.data earlier.
Added a test that stubs stage.element to avoid engine dependency.
Validates that router state is populated before component initialization.
Ensures init() hook can safely access this.$router.state.

Why

Previously, there was uncertainty whether router state was available during init(). This test guarantees state properties are injected at the correct lifecycle stage, preventing regressions.

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -220,10 +220,6 @@ export const navigate = async function () {
         ...navigationData,
         ...queryParamsData,
       }
-      state.path = route.path
-      state.params = route.params || {}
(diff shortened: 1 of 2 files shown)


michielvandergeest (approved):
LGTM

PR #470, opened 2025-09-01
Fixed triggering reactivity for all array methods

erikhaandrikman:
Small fix so we iterate over all defined array patch methods


michielvandergeest (approved):
LGTMBBQ!

PR #472, opened 2025-09-02
Updated componentId to $componentId

suresh-gangumalla:
1. Updated componentId property to $componentId
2. Removed deleting [symbols.computed] from component instance and added deleting [symbols.computedKeys] as computed symbol never exists in a component instance. we are only storing [symbols.computedKeys]


michielvandergeest (approved):
LGTM

PR #473, opened 2025-09-02
added test coverage for announcer

jfboeve:
Not a lot of test coverage we can do at the moment for the announcer. 

Maybe in the future we can improve the announcer to take a custom speech player where we can fake speak events with timers and such.

### DIFF
--- a/src/announcer/announcer.test.js
+++ b/src/announcer/announcer.test.js
@@ -17,7 +17,7 @@
 
 import test from 'tape'
 import announcer from './announcer.js'
-import { initLog } from 'src/lib/log.js'
+import { initLog } from '../lib/log.js'
 
 initLog()
 announcer.enable()


src/announcer/announcer.test.js
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import test from 'tape'
+import announcer from './announcer.js'
+import { initLog } from 'src/lib/log.js'
```
michielvandergeest: I think this import is not correct, should be something like `../../lib/log.js` I guess

PR #477, opened 2025-09-05
feat: optimize inspector-data attribute performance

il-sairamg:
## Problem
The inspector-data attribute was causing unnecessary performance overhead even when the inspector was disabled or unavailable.

## Solution
Implemented two-level optimization:

### 1. Build-time Optimization (generator.js)
- Filter out inspector-data attributes completely in production builds
- Eliminates all runtime overhead in production

### 2. Runtime Optimization (element.js)  
- Check inspector launch setting before processing
- Skip expensive object validation when inspector disabled

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -465,6 +465,11 @@ const propsTransformer = {
     }
   },
   set 'inspector-data'(v) {
+    // Skip processing if inspector is not enabled for performance optimization
+    if (!Settings.get('inspector', false)) {
+      return
+    }
+
     if (typeof v === 'object' || (isObjectString(v) === true && (v = parseToObject(v)))) {
       this.props['data'] = v
     }
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -223,6 +223,12 @@ const colorMap = {
 let textDefaults = null
 
 /**
+ * Inspector enabled flag (initialized on first use).
+ * @type {boolean|null}
+ */
+let inspectorEnabled = null
+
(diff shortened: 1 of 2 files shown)


src/engines/L3/element.js
```diff
@@ -465,6 +465,11 @@ const propsTransformer = {
     }
   },
   set 'inspector-data'(v) {
+    // Skip processing if inspector is not enabled for performance optimization
+    if (!Settings.get('inspector', false)) {
```
michielvandergeest: micro optimization, but let's change this from a `falsey` check to a `false` check (`Settings.get('inspector', false) === false`)

il-sairamg: Updated the check to use strict === false as suggested.

src/engines/L3/element.js
```diff
@@ -465,6 +465,11 @@ const propsTransformer = {
     }
   },
   set 'inspector-data'(v) {
+    // Skip processing if inspector is not enabled for performance optimization
+    if (Settings.get('inspector', false) === false) {
```
michielvandergeest: can we do the Settings.get call only once maybe? We could have a global `let` and set that in the default Element function. Much like we do with textDefaults

That should be more performant than calling `Settings.get` every time

michielvandergeest (changes requested):
looks good!

one comment with a suggested optimization

src/engines/L3/element.js
il-sairamg: Updated the code as suggested.

PR #481, opened 2025-09-09
Assign null to router state data

suresh-gangumalla:
Assigning null to router state data property before assigning actual route data. This will make sure not triggering reactivity effects of previous route component


michielvandergeest (approved):
LGTM!

PR #482, opened 2025-09-10
fix: added keepPreviousAlive as override in router.to

MarcinG98:
I have added a new option which can be passed when calling router.to(). 
When navigating this option will cache the previous page (page being navigated from) so when navigating back focus etc is remained. 
Very similar to keepAlive but is passed only as an override option when calling a .to() navigation.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -235,7 +235,19 @@ declare module '@lightningjs/blits' {
      */
     keepAlive?: true
   }
+  /**
+   * Navigation Options that can be used when navigating to a new route
+   */
+  type NavigationOverride = {
+    /**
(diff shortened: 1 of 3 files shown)


michielvandergeest:
@MarcinG98, as you already know I've taken a different approach to enable the desired behaviour. Instead of introducing an arbitrary config option during the navigation (that may be causing more confusion), it would be better to handle this in a `before()` or `beforeEach()` hook.

There you can modify the current route instance, and depending on the situation apply `keepAlive` (or reset it).

I think that's a cleaner solution that keeps this very specific behaviour there where it belongs (at the App side).

So closing this one for now. Thanks for helping me explore this feature, though 👍

PR #490, opened 2025-09-30
Handle braces in strings and comments correctly for computedprops

il-sairamg:
Fixes parsing failures in computed properties caused by braces inside strings and comments. 

Replaces naive brace counting with character-wise parsing that correctly ignores braces in `"strings"`, `// comments`, and `/* block comments */`.

Refactors parsing logic into reusable functions for better maintainability.


suresh-gangumalla (commented):
LGTM

PR #495, opened 2025-10-03
Fix early return in element property setting logic

il-sairamg:
Removed the early `return` statement in the transition handling
- Restructured the logic to use proper `if-else` branching
- Now both transition animations and regular property setting can occur as intended


michielvandergeest (approved):
Looks good! Thx

PR #496, opened 2025-10-07
Test Coverage Tuesday - Covered Component  Base Methods & Reactivity

suresh-gangumalla:


src/component/base/methods.test.js
```diff
@@ -0,0 +1,351 @@
...
+  assert.end()
+})
+
+test('Methods - Should add select, $select, and $render, shader methods', (assert) => {
+  const component = Object.defineProperties({}, { ...methods })
+  assert.equal(typeof component.select, 'function', 'should have select method')
+  assert.equal(typeof component.$select, 'function', 'should have $select method')
+  assert.equal(typeof component.shader, 'function', 'should have $shader method')
+  assert.end()
+})
```
michielvandergeest, line 28: Should we combine the above 3 tests into 1?

src/component/base/methods.test.js
```diff
@@ -0,0 +1,351 @@
...
+    assert.equal(component.lifecycle.state, 'focus', 'lifecycle state should be focus')
+
+    component.unfocus()
+    setTimeout(() => {
+      assert.equal(component[symbols.state].hasFocus, false, 'hasFocus should be false')
+      assert.equal(component.lifecycle.state, 'unfocus', 'lifecycle state should be init')
+      assert.end()
+    }, 100)
+  }, 100)
+})
```
michielvandergeest, line 121: It might be worth adding an actual focus and unfocus hook to the mocked object.
Especially to validate the order of setting `hasFocus` (i.e. set after the focus hook is called)

src/component/base/methods.test.js
```diff
@@ -0,0 +1,351 @@
...
+  assert.equal(shaderObj.props.prop2, 'fast', 'shader object should have correct prop2')
+  assert.end()
+})
+
+test('Methods - Validate removeGlobalEffects method behavior', (assert) => {
+  const component = Object.defineProperties({}, { ...methods })
+
+  const effects = []
+  component[symbols.removeGlobalEffects](effects)
+  assert.pass('removeGlobalEffects should execute without error')
```
michielvandergeest, line 276: we're not testing much here - but I understand, as it's tricky to do that right. That's good for now, but let's at least add a comment to improve on this.

src/component/base/methods.test.js
```diff
@@ -0,0 +1,351 @@
...
+  holderMock.destroyed = false
+
+  // mock children with destroy method and a counter
+  const childrenDestroyMock = function () {
+    childrenDestroyMock.count++
+  }
+  childrenDestroyMock.count = 0
+
+  // define a component with necessary properties
+  const component = Object.defineProperties(
```
michielvandergeest: should we abstract this to a test helper function (so we can reuse it in different places)?

src/lib/reactivity/reactive.test.js
```diff
@@ -25,3 +27,192 @@ test('Type', (assert) => {
...
+
+test('Reactive - Assign new object to reactive property', (assert) => {
+  const data = reactive({ nested: { foo: 'foo' }, count: 0 })
+
+  const nonReactiveData = { count: 0 }
+
+  const basicEffect = () => {
+    data.nested.foo // track nested.foo
+    data.count // track count
+    nonReactiveData.count++ // not tracked since nonReactiveData is not reactive
```
michielvandergeest: smart idea to test this via a counter. Are there any downsides to this versus using an official spy? Maybe a spy is more versatile, but don't know if we need that

If we go with this pattern, can't we just use a `let counter` ? Having an object called `nonReactiveData` threw me off a little bit in the beginning tryinf to understand this test.

src/lib/reactivity/reactive.test.js
```diff
@@ -25,3 +27,192 @@ test('Type', (assert) => {
...
+
+test('Reactive - Get raw object from proxy', (assert) => {
+  const originalObj = { a: 100, b: 200 }
+  const proxiedObj = reactive(originalObj)
+
+  assert.deepEqual(
+    proxiedObj[symbols.raw],
+    originalObj,
+    'Should return the original object from the proxy using the raw symbol'
+  )
```
michielvandergeest, line 78: for the sake of being thorough, maybe also add a test that `originalObj` does not equal `proxiedObject` ?

src/component/base/methods.test.js
suresh-gangumalla: Added actual focus and unfocus hooks

src/lib/reactivity/reactive.test.js
suresh-gangumalla: Done, moved object to simple counter.

PR #497, opened 2025-10-07
feat: improve test coverage for routes, announcer, timeouts, and utils

il-sairamg:
- Add comprehensive test coverage for routes.js
- Improve announcer.js test coverage
- Achieve full coverage for timeouts_intervals.js + fix clearInterval bug
- Improve utils.js test coverage
- Fix TypeScript errors in announcer tests

### DIFF
--- a/src/announcer/announcer.test.js
+++ b/src/announcer/announcer.test.js
@@ -38,99 +38,187 @@ test('Announcer 1 Second Pause Remove', (assert) => {
     assert.end()
   })
   assert.comment('remove pause')
-  pause['remove']()
+  pause.remove()
 })
 
 test('Announcer disable/enable/toggle', (assert) => {
+  let disabledProcessed = false
+  let enabledProcessed = false
+
+  // Test disabled - messages should NOT be processed
   announcer.disable()
-  assert.equal(typeof announcer.speak('test').then, 'function', 'Disabled returns noop')
+  const disabled1 = announcer.speak('disabled msg 1')
+  const disabled2 = announcer.speak('disabled msg 2')
 
-  announcer.toggle(true)
-  assert.equal(typeof announcer.speak('test').then, 'function', 'Toggle on works')
+  assert.equal(disabled1, disabled2, 'Disabled returns same noop object')
 
-  announcer.toggle(false)
-  assert.equal(typeof announcer.speak('test').then, 'function', 'Toggle off returns noop')
+  disabled1.then(() => {
+    disabledProcessed = true
+  })
 
+  // Test enabled - messages SHOULD be processed/queued
   announcer.enable()
-  assert.end()
+  const enabled1 = announcer.speak('enabled msg 1')
+  const enabled2 = announcer.speak('enabled msg 2')
+
+  assert.notEqual(enabled1, disabled1, 'Enabled returns real promise, not noop')
+  assert.notEqual(enabled1, enabled2, 'Each enabled message gets unique promise')
+
+  enabled1.then(() => {
+    enabledProcessed = true
+  })
+
+  // Test toggle off
+  announcer.toggle(false)
+  const toggledOff = announcer.speak('toggled off msg')
+  assert.equal(toggledOff, disabled1, 'Toggle off returns same noop')
+
+  // Test toggle on
+  announcer.toggle(true)
+  const toggledOn = announcer.speak('toggled on msg')
+  assert.notEqual(toggledOn, disabled1, 'Toggle on returns real promise')
+
+  // Verify processing behavior after a delay
+  setTimeout(() => {
+    assert.false(disabledProcessed, 'Disabled messages are NOT processed')
+    assert.ok(enabledProcessed || enabled1, 'Enabled messages are queued for processing')
+
+    // Clean up
+    enabled1.cancel()
+    enabled2.cancel()
+    toggledOn.cancel()
+
+    assert.end()
+  }, 100)
 })
 
-test('Announcer methods', (assert) => {
+test('Announcer methods exist', (assert) => {
   announcer.enable()
 
   assert.equal(typeof announcer.stop, 'function', 'Stop method exists')
   assert.equal(typeof announcer.clear, 'function', 'Clear method exists')
-  assert.equal(typeof announcer.polite('test').then, 'function', 'Polite works')
-  assert.equal(typeof announcer.assertive('test').then, 'function', 'Assertive works')
-  assert.equal(
-    typeof announcer.speak('test', 'polite').then,
-    'function',
-    'Speak with politeness works'
-  )
-  assert.equal(typeof announcer.pause(100).then, 'function', 'Pause works')
+  assert.equal(typeof announcer.polite, 'function', 'Polite method exists')
+  assert.equal(typeof announcer.assertive, 'function', 'Assertive method exists')
+  assert.equal(typeof announcer.speak, 'function', 'Speak method exists')
+  assert.equal(typeof announcer.pause, 'function', 'Pause method exists')
 
   assert.end()
+})
+
+test('Announcer politeness affects queue order', (assert) => {
+  announcer.enable()
+  const order = []
+  let completed = 0
+
+  // Queue messages with different politeness levels
+  const msg1 = announcer.speak('first', 'off')
+  const msg2 = announcer.speak('second', 'polite')
+  const msg3 = announcer.speak('third', 'assertive') // Should jump to front!
+
+  function checkComplete() {
+    completed++
+    if (completed === 3) {
+      // All three resolved, check order
+      assert.equal(order[0], 'third', 'Assertive message processes first')
+      assert.equal(order[1], 'first', 'Off message processes second')
+      assert.equal(order[2], 'second', 'Polite message processes third')
+      assert.end()
+    }
+  }
+
+  // Track resolution order
+  msg1.then(() => {
+    order.push('first')
+    checkComplete()
+  })
+  msg2.then(() => {
+    order.push('second')
+    checkComplete()
+  })
+  msg3.then(() => {
+    order.push('third')
+    checkComplete()
+  })
 })
 
 test('Announcer announcement methods', (assert) => {
   announcer.enable()
 
   const announcement = announcer.speak('test')
-  assert.equal(typeof announcement['cancel'], 'function', 'Has cancel method')
-  assert.equal(typeof announcement['remove'], 'function', 'Has remove method')
-  assert.equal(typeof announcement['stop'], 'function', 'Has stop method')
+  assert.equal(typeof announcement.cancel, 'function', 'Has cancel method')
+  assert.equal(typeof announcement.remove, 'function', 'Has remove method')
+  assert.equal(typeof announcement.stop, 'function', 'Has stop method')
 
-  // Call methods to test they work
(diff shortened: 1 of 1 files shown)


src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
     assert.end()
   })
   assert.comment('remove pause')
-  pause.remove()
+  pause['remove']()
```
michielvandergeest: why this change? it's effectively the same I know, but the current version reads a bit less obvious I think? 🙂

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+
+  announcer.toggle(true)
+  assert.equal(typeof announcer.speak('test').then, 'function', 'Toggle on works')
+
+  announcer.toggle(false)
+  assert.equal(typeof announcer.speak('test').then, 'function', 'Toggle off returns noop')
+
+  announcer.enable()
+  assert.end()
+})
```
michielvandergeest, line 93: this doesn't test much, I'd say.

Maybe we should test this a bit more thorough. Either with a spy to make sure an actual speak functions isn't called. Or by queuing some messages and ensuring they are not processed while the announcer is disabled?

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+
+  assert.equal(typeof announcer.stop, 'function', 'Stop method exists')
+  assert.equal(typeof announcer.clear, 'function', 'Clear method exists')
+  assert.equal(typeof announcer.polite('test').then, 'function', 'Polite works')
+  assert.equal(typeof announcer.assertive('test').then, 'function', 'Assertive works')
+  assert.equal(
+    typeof announcer.speak('test', 'polite').then,
+    'function',
+    'Speak with politeness works'
+  )
```
michielvandergeest: this is not a real test? we should check that, in case politeness is passed in the speak function it ends up like that in the queue as well

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+  assert.end()
+})
+
+test('Announcer announcement methods', (assert) => {
+  announcer.enable()
+
+  const announcement = announcer.speak('test')
+  assert.equal(typeof announcement['cancel'], 'function', 'Has cancel method')
+  assert.equal(typeof announcement['remove'], 'function', 'Has remove method')
+  assert.equal(typeof announcement['stop'], 'function', 'Has stop method')
```
michielvandergeest: why the array key notion? instead of the more conventional dot notation?

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+
+  // Call methods to test they work
+  announcement['cancel']()
+  announcement['stop']()
```
michielvandergeest: should we also test if they work?

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+
+  assert.end()
+})
+
+test('Announcer stop when current announcement is active', (assert) => {
+  announcer.enable()
+
+  // Create announcement and immediately stop to trigger currentId check
+  const announcement = announcer.speak('test message')
+  announcement['stop']()
```
michielvandergeest: can we test if something was actually stopped? I believe it sends an error then (interrupted, if I recall) - we should verify against that. Also, any specific reason for using array key notation?

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+  announcer.enable()
+
+  // Create multiple announcements to trigger debounce and speech synthesis
+  const announcement1 = announcer.speak('first message')
+  const announcement2 = announcer.speak('second message')
+
+  // Test that announcements are created
+  assert.equal(typeof announcement1.then, 'function', 'First announcement created')
+  assert.equal(typeof announcement2.then, 'function', 'Second announcement created')
+
```
michielvandergeest, line 181: should we also assert that  an announcement has a remove / cancel and stop method? And can we test that the announcement ended up in the queue (and maybe also a separate test that `remove` removes it from the queue)

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+})
+
+test('Announcer queue processing with debounce', (assert) => {
+  announcer.enable()
+
+  // Create multiple rapid announcements to trigger debounce logic
+  announcer.speak('message 1')
+  announcer.speak('message 2')
+  announcer.speak('message 3')
+
```
michielvandergeest, line 195: this test doesn't have any assertion 🙂

src/announcer/announcer.test.js
```diff
@@ -38,5 +38,99 @@ test('Announcer 1 Second Pause Remove', (assert) => {
...
+})
+
+test('Announcer stop current processing', (assert) => {
+  announcer.enable()
+
+  // Create announcement and stop it to trigger the stop logic
+  const announcement = announcer.speak('test message')
+
+  // Call stop to potentially trigger if currentId matches
+  announcement['stop']()
```
michielvandergeest: again we don't actually test anything here ..

src/component/base/timeouts_intervals.js
```diff
@@ -78,7 +78,7 @@ export default {
       const index = this[symbols.intervals].indexOf(intervalId)
       if (index > -1) {
         this[symbols.intervals].splice(index, 1)
-        clearTimeout(intervalId)
+        clearInterval(intervalId)
```
michielvandergeest, line 81: great find! ❤️

src/component/base/timeouts_intervals.test.js
```diff
@@ -0,0 +1,170 @@
...
+    eol: false,
+    [symbols.timeouts]: [],
+  }
+
+  const timeoutId = timeoutsIntervals.$setTimeout.value.call(component, () => {}, 100)
+  assert.ok(timeoutId !== undefined, 'Should return timeout ID')
+  assert.equal(component[symbols.timeouts].length, 1, 'Should add timeout to array')
+
+  // Clear the timeout immediately to prevent it from executing
+  timeoutsIntervals.$clearTimeout.value.call(component, timeoutId)
```
michielvandergeest: just for the sake of it I would clear it with a regular clearTimeout (and leave the `$clearTimeout` coverage only in dedicated tests)

src/component/base/timeouts_intervals.test.js
```diff
@@ -0,0 +1,170 @@
...
+  assert.end()
+})
+
+test('$clearTimeout method', (assert) => {
+  const component = { [symbols.timeouts]: [123, 456] }
+
+  timeoutsIntervals.$clearTimeout.value.call(component, 123)
+  assert.equal(component[symbols.timeouts].length, 1, 'Should remove timeout from array')
+  assert.equal(component[symbols.timeouts][0], 456, 'Should keep other timeouts')
+
```
michielvandergeest, line 70: should we actually also test if the` $clearTimeout` method actually really clears a timeout by actually scheduling one and ensuring that it's not called? Maybe we can set a value to an outer scope variable in the timeout that we schedule and make sure that it's not changed (i.e. the timeout is not called because it's cleared)

src/component/base/timeouts_intervals.test.js
```diff
@@ -0,0 +1,170 @@
...
+  assert.end()
+})
+
+test('$clearTimeouts method', (assert) => {
+  const component = { [symbols.timeouts]: [123, 456, 789] }
+  timeoutsIntervals.$clearTimeouts.value.call(component)
+  assert.equal(component[symbols.timeouts].length, 0, 'Should clear all timeouts')
+
```
michielvandergeest, line 97: Same comment as above. God test to see if the array is empty. But let's also test if the clearTimeout functionality is called

src/component/base/timeouts_intervals.test.js
```diff
@@ -0,0 +1,170 @@
...
+    eol: false,
+    [symbols.intervals]: [],
+  }
+
+  const intervalId = timeoutsIntervals.$setInterval.value.call(component, () => {}, 100)
+  assert.ok(intervalId !== undefined, 'Should return interval ID')
+  assert.equal(component[symbols.intervals].length, 1, 'Should add interval to array')
+
+  // Clear the interval immediately to prevent it from running
+  timeoutsIntervals.$clearInterval.value.call(component, intervalId)
```
michielvandergeest: same comment as for clearTimeout

src/component/base/timeouts_intervals.test.js
```diff
@@ -0,0 +1,170 @@
...
+  assert.end()
+})
+
+test('$clearInterval method', (assert) => {
+  const component = { [symbols.intervals]: [123, 456] }
+
+  timeoutsIntervals.$clearInterval.value.call(component, 123)
+  assert.equal(component[symbols.intervals].length, 1, 'Should remove interval from array')
+  assert.equal(component[symbols.intervals][0], 456, 'Should keep other intervals')
+
```
michielvandergeest, line 150: same comments as for timeouts

src/component/base/timeouts_intervals.test.js
```diff
@@ -0,0 +1,170 @@
...
+  assert.equal(component[symbols.timeouts].length, 3, 'Should store multiple timeouts')
+  assert.equal(component[symbols.timeouts][0], timeoutId1, 'Should store first timeout')
+  assert.equal(component[symbols.timeouts][1], timeoutId2, 'Should store second timeout')
+  assert.equal(component[symbols.timeouts][2], timeoutId3, 'Should store third timeout')
+
+  // Clear one timeout
+  timeoutsIntervals.$clearTimeout.value.call(component, timeoutId2)
+  assert.equal(component[symbols.timeouts].length, 2, 'Should remove only the specified timeout')
+  assert.equal(component[symbols.timeouts][0], timeoutId1, 'Should keep first timeout')
+  assert.equal(component[symbols.timeouts][1], timeoutId3, 'Should keep third timeout')
```
michielvandergeest, line 206: nice test!

But let's also add the actual clearing of intervals

src/component/setup/routes.test.js
```diff
@@ -0,0 +1,61 @@
...
+    component[symbols.routes][0].options,
+    { inHistory: true },
+    'Should apply default options when none specified'
+  )
+  assert.deepEqual(
+    component[symbols.routes][1].options,
+    { inHistory: false },
+    'Should override default options when specified'
+  )
+  assert.end()
```
michielvandergeest, line 33: nice test! 🙂

src/announcer/announcer.test.js
il-sairamg: You're right! I had used bracket notation to suppress TypeScript linting errors, but dot notation is more readable. I've reverted to dot notation.

PR #501, opened 2025-10-16
Release 1.41.0

michielvandergeest:
- Improved overall test coverage to 86%
- Filtered out inspector-data entirely in production
- Improved vite plugin that handles computed props to be more robust
- Fixed issue with global reactive object that are deeply nested
- Bumped MSDF generator dependency to address security issues in dependencies


suresh-gangumalla (approved):
LGTM

PR #505, opened 2025-10-17
Fix - Router navigation cancel on error or rejected "before" hook

ruggero-nardi:
### Code change description:

- Now when a "before" hook is executed, this is done inside a try-catch. This way the before hook will behave exactly as before when executed succesfully, and now it will catch errors and promise rejections.
- In the catch execution a flag is set to true to prevent a navigate function execution, this way the browser history can be reverted.
- the currentRoute is set back to the "previousRoute" to prevent incorrect future back() navigations.

### Bug description:

When a route is set to do checks before navigating, the "before" hook works incorrectly when trying to prevent navigation.

- True/false returned:
     - Documentation states that a boolean can be returned, but this is not possible and won't have any effect in the code.
- String or Route returned:
     - OK
- Promise<string|Route> returned:
     - If resolve(to): Navigates away to the new route.
     - If resolve(from): Navigates to the same current route (Refreshes the current page) 
     - **If reject(): Error, app stucked**
     - **If resolve(undefinedPath): Navigation prevented, Blits error shown in console.**

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -78,8 +78,9 @@ let navigationData = {}
 let navigatingBack = false
 let navigatingBackTo = undefined
 let previousFocus
-let preventHashChangeNavigation = false // Skips internal router navigation when set to true only for the next "navigate" execution, needed for window.history management
-
+// Skips internal router navigation when set to true only for the next "navigate"
+// execution, needed for window.history management
+let preventHashChangeNavigation = false
 /**
  * Get the current hash
  * @returns {Hash}
@@ -264,7 +265,7 @@ export const navigate = async function () {
               return
             }
           } catch (error) {
-            console.error('Error or Rejected Promise in "BeforeEach" Hooks', error)
+            Log.error('Error or Rejected Promise in "BeforeEach" Hook', error)
 
             preventHashChangeNavigation = true
             currentRoute = previousRoute
@@ -303,7 +304,7 @@ export const navigate = async function () {
             return
           }
         } catch (error) {
-          console.error('Error or Rejected Promise in "Before" Hook', error)
+          Log.error('Error or Rejected Promise in "Before" Hook', error)
 
           preventHashChangeNavigation = true
           currentRoute = previousRoute


src/components/RouterView.js
```diff
@@ -30,14 +30,21 @@ export default () =>
...
         if (this.parent[symbols.routerHooks] && this.parent[symbols.routerHooks].init) {
           await this.parent[symbols.routerHooks].init.apply(this.parent)
         }
-        hashchangeHandler = () => Router.navigate.apply(this)
+        hashchangeHandler = () => {
+          if (this[symbols.state].preventHashChangeNavigation) {
+            this[symbols.state].preventHashChangeNavigation = false
+            return
+          }
+          Router.navigate.apply(this)
```
michielvandergeest: instead of adding an extra state variable (`preventHashChangeNavigation`) to the router view, wouldn't it be more clean to keep this contained in the router logic itself?

in that case we would just always call `Router.navigate.apply(this)` and check a prevent has change variable in the router in the `navigate` method

src/router/router.js
```diff
@@ -267,10 +267,22 @@ export const navigate = async function () {
...
       if (route.hooks.before) {
-        beforeHookOutput = await route.hooks.before.call(this.parent, route, previousRoute)
-        if (isString(beforeHookOutput)) {
+        try {
+          beforeHookOutput = await route.hooks.before.call(this.parent, route, previousRoute)
+          if (isString(beforeHookOutput)) {
+            currentRoute = previousRoute
+            to(beforeHookOutput)
+            return
+          }
```
michielvandergeest, line 305: Some changes to this portion have been made by me around the same time as your PR (see: https://github.com/lightning-js/blits/pull/502/files). We should make sure that we keep that  functionality in tact

ruggero-nardi: Checked this, updated my dev branch with upstream dev branch and now it has all your changes plus the the additions to the router hooks

michielvandergeest (approved):
LGTM! Thanks for contributing this @ruggero-nardi !

PR #510, opened 2025-10-22
feat: add effects prop support to Sprite component (V2)

il-sairamg:
Convert props from array to object notation (V2 format)
- Add effects prop to Sprite component props object
- Pass effects prop to Element in template and added safety checks for required props
- Enables visual effects to be applied to sprite elements.

### DIFF
--- a/src/components/Sprite.js
+++ b/src/components/Sprite.js
@@ -22,9 +22,15 @@ import symbols from '../lib/symbols.js'
 export default () =>
   Component('Sprite', {
     template: `
-      <Element w="100%" h="100%" :texture="$texture" :color="$color" />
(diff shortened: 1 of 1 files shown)


src/components/Sprite.js
```diff
@@ -22,16 +22,26 @@ import symbols from '../lib/symbols.js'
 export default () =>
   Component('Sprite', {
     template: `
-      <Element w="100%" h="100%" :texture="$texture" :color="$color" />
+      <Element w="100%" h="100%" :texture="$texture" :color="$color" :effects="$effects" />
```
michielvandergeest, line 25: should `effects` be reactive or not, since effects can't be updated runtime now anyway?

PR #511, opened 2025-10-24
Release/1.42.0

il-sairamg:
- Added effects prop support to Sprite component
- Added support for redirecting in before hooks returning a route object (with updated path)
- Fixed meta options on routes functionality
- Enhanced router hook test coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,15 @@
 # Changelog
 
-## v1.41.1
+## v1.42.0
 
-_Current Release_
+_24 Oct 2025_
+
 
 - Added effects prop support to Sprite component
 - Added support for redirecting in before hooks returning a route object (with updated path)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## v1.41.1
+
+_Current Release_
```
michielvandergeest: we can remove this line here, we don't need it

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lightningjs/blits",
-  "version": "1.41.0",
+  "version": "1.41.1",
```
michielvandergeest: I think we should go for 1.42.0 (a minor bump) given we haven't only fixed stuff (path), but also added functionality (redirect in router)

PR #514, opened 2025-10-28
Feature/sprite reactive image prop

il-sairamg:
# feat: make Sprite image prop reactive

- Added reactive functionality to Sprite component's `image` prop
- Implemented smart texture caching for better performance

### DIFF
--- a/src/components/Sprite.js
+++ b/src/components/Sprite.js
@@ -22,7 +22,7 @@ import symbols from '../lib/symbols.js'
 export default () =>
   Component('Sprite', {
     template: `
-      <Element w="100%" h="100%" :texture="$texture" :color="$color" :effects="$effects" />
+      <Element w="100%" h="100%" :texture="$texture" :color="$color" effects="$effects" />
     `,
     props: ['image', 'map', 'frame', 'color', 'effects'],
 


src/components/Sprite.js
```diff
@@ -25,34 +25,60 @@ export default () =>
...
     computed: {
       texture() {
-        const options =
-          'frames' in this.map
-            ? Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
-            : this.map[this.frame]
+        // If no image - nothing to render
+        if (!this.image) return null
+
+        // ensure renderer is available
```
suresh-gangumalla: I'm not in favour of validating renderer here (inside  computed || inside a component) but wait for Michiel comment.

src/components/Sprite.js
```diff
@@ -25,34 +25,60 @@ export default () =>
...
-      },
-    },
-    hooks: {
-      ready() {
-        this.spriteTexture = this[symbols.renderer]().createTexture('ImageTexture', {
-          src: this.image,
-        })
+
+        // Default fallback (single image)
+        return this.spriteTexture
```
michielvandergeest, line 92: why would the ready hook return something?

src/components/Sprite.js
```diff
@@ -25,34 +25,60 @@ export default () =>
...
+        if (!this.image) return null
+
+        // ensure renderer is available
+        const renderer = this[symbols.renderer]()
+        if (!renderer || !renderer.createTexture) {
+          return null
+        }
+
+        //  Recreate only when image src changes
+        if (!this.spriteTexture || this.currentSrc !== this.image) {
```
michielvandergeest: instead of a falsy check it would be better to do an undefined check here

src/components/Sprite.js
```diff
@@ -25,34 +25,60 @@ export default () =>
...
-            ? Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
-            : this.map[this.frame]
+        // If no image - nothing to render
+        if (!this.image) return null
+
+        // ensure renderer is available
+        const renderer = this[symbols.renderer]()
+        if (!renderer || !renderer.createTexture) {
+          return null
+        }
```
michielvandergeest, line 47: is there a valid use-case that the renderer wouldn't be available yet?
if so perhaps we should do a null check, but assign to a variable once it is, to not repeat doing `this[symbols.renderer]()` every computed execution. Also then let's move it up in the function

src/components/Sprite.js
```diff
@@ -25,34 +25,60 @@ export default () =>
...
+        let options = null
+        if (this.map && this.frame) {
+          options =
+            'frames' in this.map
+              ? Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
+              : this.map[this.frame]
+        }
+
+        //  If no map but frame is object (manual subtexture)
+        if (!options && typeof this.frame === 'object') {
```
michielvandergeest: undefined check would be better than falsy check

src/components/Sprite.js
```diff
@@ -25,34 +25,60 @@ export default () =>
...
+          options = this.frame
+        }
+
+        // Create SubTexture only if frame data exists
+        if (options) {
```
michielvandergeest: let's make this a not undefined check

michielvandergeest (changes requested):
Nice work so far!

Left some coments

src/components/Sprite.js
il-sairamg: component doesn’t have a ready hook. This return is required for the computed property to work with the template binding.

michielvandergeest: Sorry, I got confused with the diff.

PR #517, opened 2025-10-29
Replaced 'assert' with 'with' in launch.js to fix incompatibility issue with modern node versions

gasproni:
The file src/launch.js made use of the 'assert' keyword at line 21: 

`import blitsPackageInfo from '../package.json' assert { type: 'json' }`

"assert" is now deprecated, and needs to be replaced with "with". That import causes issues when running unit tests (it probably causes problems to other applications too).


michielvandergeest (approved):
LGTM

gasproni:
Hi, first a disclaimer: I'm far from being a JS or Typescript expert. I did some research and found out that the first version of node with support for "with" is 18.20.0, then the support for "assert" was removed in 22.0.0. Apparently, but I haven't had the time to test it, another possibility may be to run "npx codemod run @nodejs/import-assertions-to-attributes" to do a migration for modern versions of node. I guess the choice of what to do depends on the amount of backward compatibility you want to provide.

PR #519, opened 2025-10-31
prevent infinite recursion for :effects on components

il-sairamg:
Fixes infinite recursion when using `:effects` with reactive values on components.

- Skip reactive updates for `:effects` on components
- Fix array parsing for `effects` (without colon) on components

### DIFF
--- a/src/lib/codegenerator/generator.js
+++ b/src/lib/codegenerator/generator.js
@@ -312,11 +312,14 @@ const generateComponentCode = function (
 
   Object.keys(templateObject).forEach((key) => {
     if (isReactiveKey(key)) {
-      this.effectsCode.push(`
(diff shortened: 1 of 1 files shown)


src/lib/codegenerator/generator.js
```diff
@@ -312,11 +312,14 @@ const generateComponentCode = function (
 
   Object.keys(templateObject).forEach((key) => {
     if (isReactiveKey(key)) {
-      this.effectsCode.push(`
-        ${elm}[Symbol.for('props')]['${key.substring(1)}'] = ${interpolate(
-          templateObject[key],
-          options.component
-        )}`)
+      // Skip reactivity for :effects on components to prevent infinite recursion
```
suresh-gangumalla, line 315: @il-sairamg I believe, We planned to remove `:` from Sprite component not from code generator.

PR #520, opened 2025-11-04
Moved ESlint to a separate script

suresh-gangumalla:
Observed that eslint and eslint fix npm run commands are not working in windows machine. We need to escape double quotes to make it work in Windows OS. I have moved complete eslint operations to independent script which list all js files using global pattern.

Additionally, 
- Removed eslintrc.cjs which is no logner needed after migrating to .esline.config.cjs
- Removed eslintignore file, which is not longer needed as we are configuring ignore files in Global ignore of configuration file (removing this file clears the warning message by eslint)

### DIFF
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -7,7 +7,6 @@ const __filename = fileURLToPath(import.meta.url)
 const __dirname = resolve(__filename, '../..')
 
 // Get "fix" flag from command line arguments
-console.log(process.argv)
 const isFix = process.argv[2] === 'fix'
 
 console.log('Starting the linting process...')


scripts/lint.js
```diff
@@ -0,0 +1,51 @@
+import fg from 'fast-glob'
+import { resolve } from 'path'
+import { fileURLToPath } from 'url'
+import { ESLint } from 'eslint'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = resolve(__filename, '../..')
+
+// Get "fix" flag from command line arguments
+console.log(process.argv)
```
michielvandergeest: left over console.log?

suresh-gangumalla: Yes, I will remove and push an update @michielvandergeest

PR #521, opened 2025-11-05
feat(announcer): add utterance options and global config

il-sairamg:
Added support for per-call utterance options (rate, pitch, lang, voice, volume).
Introduced a new configure() method to define global default settings for all speech announcements.
Integrated the configuration with Blits settings for consistent behavior across components. 
Ensured backward compatibility — existing announcer usage remains unaffected.

### DIFF
--- a/src/announcer/announcer.js
+++ b/src/announcer/announcer.js
@@ -127,22 +127,13 @@ const processQueue = async () => {
     debounce = setTimeout(() => {
       Log.debug(`Announcer - speaking: "${message}" (id: ${id})`)
 
-      // Merge global defaults with per-call options
-      const mergedOptions = { ...globalDefaultOptions, ...options }
-
-      // Extract utterance options from merged options object
-      const utteranceOptions = {
-        message,
-        id,
-        lang: mergedOptions.lang,
-        pitch: mergedOptions.pitch,
-        rate: mergedOptions.rate,
-        voice: mergedOptions.voice,
-        volume: mergedOptions.volume,
-      }
-
       speechSynthesis
-        .speak(utteranceOptions)
+        .speak({
+          message,
+          id,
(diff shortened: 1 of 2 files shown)


src/announcer/announcer.js
```diff
@@ -124,8 +127,22 @@ const processQueue = async () => {
...
+      const utteranceOptions = {
+        message,
+        id,
+        lang: mergedOptions.lang,
+        pitch: mergedOptions.pitch,
+        rate: mergedOptions.rate,
+        voice: mergedOptions.voice,
+        volume: mergedOptions.volume,
+      }
+
```
michielvandergeest: we could probably do this is a single spread? that would save us from creating and allocating 2 variables over one.

If it still reads well, we could also consider doing it inline in the `speak` function on L145

src/announcer/announcer.js
```diff
@@ -159,6 +176,10 @@ const clear = () => {
   queue.length = 0
 }
 
+const configure = (options = {}) => {
+  globalDefaultOptions = { ...globalDefaultOptions, ...options }
```
michielvandergeest, line 171: do we need to spread the globalDefaultOptions (given it's an empty object?) Or can we just assign straight?

michielvandergeest (changes requested):
This looks good! I added 2 small comments.

Also we may need to add a paragraph of documentation on the new configure method and the options param added to the speak function

src/announcer/announcer.js
il-sairamg: We use the spread operator to merge new options with existing ones; without it, later configure() calls would overwrite the previous settings

michielvandergeest: okay, that makes sense

PR #523, opened 2025-11-11
Common vite dev config plugin to define anything

suresh-gangumalla:
This plugin provides a central place in the Blits framework to manage dev-only Vite settings. It helps keep app configs clean and makes it easier to handle updates or deprecations in Vite over time.

### DIFF
--- a/vite/index.js
+++ b/vite/index.js
@@ -28,9 +28,9 @@ export { default as reactivityGuard } from './reactivityGuard.js'
 export { default as injectDevConfig } from './inject-dev-config.js'
 
 export default [
+  injectDevConfig(),
   blitsFileConverter(),
   reactivityGuard(),
   preCompiler(),
   msdfGenerator(),
-  injectDevConfig(),
 ]


michielvandergeest:
will this overwrite user config?

vite/index.js
```diff
@@ -19,10 +19,18 @@ import preCompiler from './preCompiler.js'
...
 export { default as reactivityGuard } from './reactivityGuard.js'
+export { default as injectDevConfig } from './inject-dev-config.js'
 
-export default [blitsFileConverter(), reactivityGuard(), preCompiler(), msdfGenerator()]
+export default [
+  blitsFileConverter(),
+  reactivityGuard(),
+  preCompiler(),
+  msdfGenerator(),
+  injectDevConfig(),
```
michielvandergeest: just thinking we maybe should move this to the first position? then all the other plugins could potentially take advantage of the __DEV__ flag as well (or at least be aligned across the chain)

PR #528, opened 2025-11-12
chore: add $clearListeners API and add $unlisten type definition

tuncayuk:
**Summary**

This PR introduces;

- $clearListeners API: Removes all event listeners for the current component from all events.

- Add type definition for the $unlisten API

### DIFF
--- a/docs/components/utility_methods.md
+++ b/docs/components/utility_methods.md
@@ -248,6 +248,8 @@ export default Blits.Component('MyComponent', {
 
 The `$clearListeners` utility method (available on each Blits Component) is designed to remove all event listeners registered on the component via $listen(). This method is useful when you want to ensure that the component stops listening to all events at once, rather than removing them individually with `$unlisten()`.
 
+> Note: there is no need to manually remove listeners when destroying a component. The Blits teardown mechanism will ensure that listeners are removed in this case.
+
 ```js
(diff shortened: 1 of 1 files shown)


src/component/base/events.js
```diff
@@ -72,4 +72,18 @@ export default {
     enumerable: true,
     configurable: false,
   },
+
+  /**
+   * Removes all listeners for this component from all events.
+   *
+   * @returns {void}
+   */
+  $clearListeners: {
```
michielvandergeest, line 81: Nice addition! Can we add this method to the documentation at `docs/components/utility_methods.md` ?

index.d.ts
```diff
@@ -375,6 +383,13 @@ declare module '@lightningjs/blits' {
      */
     $emit(name: string, data?: any, byReference?: boolean): void;
 
+    /**
+     * Remove all listeners for this component from all events
```
michielvandergeest, line 387: shouldn't "for this component from all events" be "from this component for all events" ?

src/component/base/events.js
tuncayuk: I have added docs for the `$clearListeners`. Thanks!

index.d.ts
tuncayyildirtanuk: Changed. Thanks!

PR #532, opened 2025-11-13
fix: only navigate back on before hook error if history exits

MarcinG98:
If an error occurs on the home route, currently the router will call window.history.back. If there is no history it will just keep firing the home route over and over again in a loop. 

This ensures that we only perform these navigation rules if there is history.


michielvandergeest (approved):
lgtm!

PR #534, opened 2025-11-17
Introduce $bubbleInput() to forward key events up the component tree

il-sairamg:
Added $bubbleInput() to bubble keyboard events to parent components without changing focus.
Mirrors existing input-handling behavior for consistent event propagation.
Safely registers key-up callbacks only when returned by handlers.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -27,7 +27,7 @@ import { DEFAULT_HOLD_TIMEOUT_MS, DEFAULT_KEYMAP } from './constants.js'
  * Initialized once during application init.
  * @type {Object<string, string>}
  */
-export let keyMap = { ...DEFAULT_KEYMAP }
+export let keyMap = {}
 
 const Application = (config) => {
   config.hooks = config.hooks || {}


src/component/base/methods.js
```diff
@@ -51,6 +52,39 @@ export default {
...
+  $bubbleInput: {
+    /**
+     * Bubble up a keyboard event to a parent component without changing focus
+     * @this {import('../../component').BlitsComponent}
+     * @param {string} key - The key name (e.g., 'enter', 'back', 'up')
+     * @param {KeyboardEvent} event - The keyboard event to bubble up
+     * @returns {boolean} - Returns true if a parent component handled the event, false otherwise
+     */
+    value: function (key, event) {
+      if (!this.parent) return false
```
michielvandergeest: better to do an `undefined` check than a generic `falsy` check

src/component/base/methods.js
```diff
@@ -51,6 +52,39 @@ export default {
     enumerable: true,
     configurable: false,
   },
+  $bubbleInput: {
```
michielvandergeest: I'm not sure I like the name `$bubbleInput` ...

Maybe it should just be `$input`? Also, would it then make sense that you call this on the actual component that you want to apply the `$input` on, instead of automatically on the parent? That would make it more flexible and work pretty much the same in terms of logic as `$focus()` ..

src/component/base/methods.js
```diff
@@ -51,6 +52,39 @@ export default {
...
+    /**
+     * Bubble up a keyboard event to a parent component without changing focus
+     * @this {import('../../component').BlitsComponent}
+     * @param {string} key - The key name (e.g., 'enter', 'back', 'up')
```
michielvandergeest: is this needed as a separate param? why can't we just pass the `event` in the same way we do with `$focus()` ?

src/component/base/methods.js
```diff
@@ -23,6 +23,26 @@ import { trigger } from '../../lib/reactivity/effect.js'
...
+  Escape: 'escape',
+  37: 'left',
+  39: 'right',
+  38: 'up',
+  40: 'down',
+  13: 'enter',
+  32: 'space',
+  8: 'back',
+  27: 'escape',
+}
```
michielvandergeest: this default keymap also exists in `application.js` already - we probably shouldn't duplicate it, but reuse it

src/component/base/methods.js
```diff
@@ -52,21 +72,24 @@ export default {
...
+     * @param {KeyboardEvent} event - The keyboard event to handle
+     * @returns {boolean} - Returns true if this component or a parent component handled the event, false otherwise
      */
-    value: function (key, event) {
-      if (!this.parent) return false
+    value: function (event) {
+      if (event === null || event === undefined || event instanceof KeyboardEvent === false)
+        return false
+
+      const keyMap = { ...defaultKeyMap, ...Settings.get('keymap', {}) }
```
michielvandergeest: we don't have to do this spreading and getting settings every $input invocation right?
maybe we should do this only once (in application.js? or abstracted in a helper which application.js also uses) and export it from there so we have one single place to generate the key map to use in the app

docs/components/user_input.md
```diff
@@ -59,9 +59,9 @@ When a component handles a key press by having a corresponding function specifie
 
 1. **Change focus only (no bubbling)**: Use `this.parent.$focus()` without the event parameter. This only changes focus to the parent and does not bubble the event.
 2. **Change focus and bubble the event**: Use `this.parent.$focus(e)` with the event parameter. This changes focus to the parent AND bubbles the event to the parent's input handler.
-3. **Bubble event without changing focus**: Use `this.$bubbleInput(key, e)`. This bubbles the event to the parent's input handler WITHOUT changing focus.
+3. **Handle input without changing focus**: Use `this.parent.$input(e)`. This handles the input event on the parent WITHOUT changing focus.
```
michielvandergeest: technically we can also use this to `$select` a component and execute `$input()` on that one. `this.parent` will be the most common use case, but could be valuable to add

docs/components/user_input.md
```diff
@@ -59,9 +59,9 @@ When a component handles a key press by having a corresponding function specifie
...
 2. **Change focus and bubble the event**: Use `this.parent.$focus(e)` with the event parameter. This changes focus to the parent AND bubbles the event to the parent's input handler.
-3. **Handle input without changing focus**: Use `this.parent.$input(e)`. This handles the input event on the parent WITHOUT changing focus.
+3. **Handle input without changing focus**: Use `this.$input(e, this.parent)`. This handles the input event on the parent WITHOUT changing focus. You can also use `this.$input(e, this.$select('ref'))` to handle input on any component selected by ref.
 
-**Note**: Both `this.parent.$focus(e)` and `this.parent.$input(e)` process events on the parent. The difference is that `$focus()` changes focus, while `$input()` only handles the input event.
+**Note**: Both `this.parent.$focus(e)` and `this.$input(e, this.parent)` process events on the parent. The difference is that `$focus()` changes focus, while `$input()` only handles the input event. The `$input()` method accepts an optional second parameter to specify which component should handle the input.
```
michielvandergeest: No, this should work like this:

```
this.$select('componentByRef').$input(e)
```

So the idea is that we execute the `$input` method on a specific component.

docs/components/utility_methods.md
```diff
@@ -129,22 +131,29 @@ export default Blits.Component('MyComponent', {
...
         return
       }
       // No parent handler, handle it ourselves
       this.handleBack()
     },
+    space(e) {
+      // Handle input on a component selected by ref
+      const menu = this.$select('mainMenu')
+      if (menu) {
+        this.$input(e, menu)
```
michielvandergeest: so in this case the example should be

```
const menu = this.$select('mainMenu')
if (menu) {
    menu.$input(e)
 }
```

src/application.js
```diff
@@ -20,28 +20,28 @@ import { default as Focus, keyUpCallbacks } from './focus.js'
...
+ * Get the merged keyMap (default + custom settings).
+ * Computed once and cached for performance.
+ * @returns {Object<string, string>} The merged keyMap
+ */
+export const getKeyMap = () => {
+  if (keyMap === null) {
+    keyMap = { ...DEFAULT_KEYMAP, ...Settings.get('keymap', {}) }
   }
+  return keyMap
+}
```
michielvandergeest: this pattern still means that we're executing this function and doing the settings fetch and spread every time.

wouldn't it be better to export the `let keyMap` to be used in different portions of the app and execute the logic inside this function only once during initialization?

src/application.js
```diff
@@ -20,28 +20,16 @@ import { default as Focus, keyUpCallbacks } from './focus.js'
...
-    32: 'space',
-    8: 'back',
-    27: 'escape',
-  }
+/**
+ * Merged keyMap (default + custom settings).
+ * Initialized once during application init.
+ * @type {Object<string, string>}
+ */
+export let keyMap = { ...DEFAULT_KEYMAP }
```
michielvandergeest: this can be just an empty object I think?

il-sairamg: @michielvandergeest   Updated with empty object.

michielvandergeest (approved):
LGTM!

PR #544, opened 2025-11-26
fix: resumeInfinity/onresume interaction

tuncayuk:
1. Change resumeInfinity to reset the existing timeout instead of calling clear() when a timer is present

2. Introduce resumeFromKeepAlive guard so keep-alive resume() calls don’t recursively trigger resumeInfinity

# Key changes for consumers:

## 1) Added enableUtteranceKeepAlive parameter

Previously, the startKeepAlive() logic (resume/pause) was enabled only for non-Android user agents via:

```js
const isAndroid = /android/i.test((window.navigator || {}).userAgent || '')
```

However, on Comcast XI devices, triggering SpeechSynthesisUtterance resume/pause events can cause the announce to get stuck (the Promise never resolves). Also, for Comcast XI we don’t need to run the startKeepAlive() (resume/pause) logic.

With the new enableUtteranceKeepAlive option, consumers can explicitly enable/disable the keep-alive behavior from the app side:

```js
this.$announcer.configure({ enableUtteranceKeepAlive: false });
this.$announcer.enable();
```

Note: The default value is based on defaultUtteranceKeepAlive, so consumers can override it if needed. This is not a breaking change.
const defaultUtteranceKeepAlive = !isAndroid

## 2) Added cancelPrevious parameter

While working with Vishnu, I noticed they were cancelling the previous utterance during navigation by calling three APIs:
```js
this.$announcer.stop();
this.$announcer.clear();
this.$announcer.speak(utterance);
```

To simplify this requirement, I added cancelPrevious, which cancels any in-progress/queued utterance before speaking the new one:

```js
this.$announcer.speak(utterance, 'off', { cancelPrevious: true });
```

This reduces multiple calls into a single, safer API call.

## 3) Improved clearing of utterances

I improved how we clear utterances by removing them based on the utterance ID, ensuring references are cleaned up more reliably and avoiding stale entries in the map.

### DIFF
--- a/docs/plugins/text-to-speech-announcer.md
+++ b/docs/plugins/text-to-speech-announcer.md
@@ -49,7 +49,7 @@ Blits.Component('MyComponent', {
 
 ### Execute logic when done
 
-The `speak()`-function returns a promise that resolves when it's done speaking, or when it's interupted or errors out. The reason of the `speak()` method resolving is passed as an argument into the resolve function.
+The `speak()`-function returns a promise that resolves when it's done speaking, or when it's interrupted or errors out. The reason of the `speak()` method resolving is passed as an argument into the resolve function.
 
 ```js
 this.$announcer.speak('Hello').then((reason) => console.log('Speaking hello finished', reason))
--- a/src/announcer/announcer.test.js
+++ b/src/announcer/announcer.test.js
@@ -171,9 +171,9 @@ test('Announcer stop interrupts processing', (assert) => {
   const announcement = announcer.speak('test message for interruption')
 
   announcement.then((status) => {
-    // Should resolve with 'interupted' when stop() is called
+    // Should resolve with 'interrupted' when stop() is called
     assert.ok(
-      status === 'interupted' || status === 'unavailable',
+      status === 'interrupted' || status === 'unavailable',
       'Stop interrupts (or unavailable if no speechSynthesis)'
     )
     assert.end()


src/announcer/speechSynthesis.js
```diff
@@ -33,13 +33,24 @@ const clear = () => {
   }
 }
 
+let resumeFromKeepAlive = false
+
 const resumeInfinity = (target) => {
-  if (!target || infinityTimer) {
+  // If the utterance is gone, just stop the keep-alive loop.
+  if (!target) {
```
michielvandergeest: can we make this into a specific condition check instead of a generic `falsy` check?

src/announcer/speechSynthesis.js
```diff
@@ -33,13 +33,24 @@ const clear = () => {
...
     return clear()
   }
 
+  // We only ever want ONE keep-alive timer running per utterance.
+  // If there's an existing timer, cancel it and start a fresh one below.
+  if (infinityTimer) {
```
michielvandergeest: instead of a generic `trythy`-check, let's make this in a "not undefined" or "not null"-check

src/announcer/speechSynthesis.js
```diff
@@ -78,6 +89,13 @@ const speak = (options) => {
     }
 
     utterance.onresume = () => {
+      // Ignore resume events that we *know* came from our own keep-alive (the pause()/resume() in resumeInfinity).
+      if (resumeFromKeepAlive) {
```
michielvandergeest: `if(resumeFromKeepAlive === true)`

src/announcer/speechSynthesis.js
tuncayuk: Done

src/announcer/speechSynthesis.js
tuncayuk: Done

src/announcer/speechSynthesis.js
tuncayuk: Done

src/announcer/announcer.js
```diff
@@ -134,37 +148,86 @@ const processQueue = async () => {
...
           currentId = null
+          currentResolveFn = null
           isProcessing = false
           Log.debug(`Announcer - error ("${e.error}") while speaking: "${message}" (id: ${id})`)
           resolveFn(e.error)
           processQueue()
         })
       debounce = null
-    }, 200)
+    }, 300)
```
michielvandergeest, line 160: Was there a specific reason for increasing the timeout from 200 to 300?

src/announcer/announcer.js
```diff
@@ -138,33 +149,77 @@ const processQueue = async () => {
...
+    resolveFn('interrupted')
+  }
 }
 
 const clear = () => {
+  // Clear debounce timer
+  clearDebounceTimer()
+
+  // Cancel any active speech synthesis
+  speechSynthesis.cancel()
```
michielvandergeest: does clearing the queue _also_ mean that we want to cancel the current message being spoken out? I'm not sure that it should always be so ..

src/announcer/speechSynthesis.js
```diff
@@ -19,33 +19,69 @@ import { Log } from '../lib/log.js'
...
-const clear = () => {
-  if (infinityTimer) {
-    clearTimeout(infinityTimer)
-    infinityTimer = null
+const clear = (id) => {
+  const state = utterances.get(id)
+  if (!state) {
+    return
+  }
+  if (state?.timer !== null) {
```
michielvandergeest: I prefer not to use optional chaining (as it doesn't transpile to very optimal es5 code)

src/announcer/speechSynthesis.js
```diff
@@ -19,33 +19,69 @@ import { Log } from '../lib/log.js'
...
+
+  // utterance status: utterance was removed (cancelled or finished)
+  if (!state) {
+    return
+  }
+
+  const { utterance } = state
+
+  // utterance check: utterance instance is invalid
+  if (!(utterance instanceof SpeechSynthesisUtterance)) {
```
michielvandergeest: do we need this instanceof check? would there ever be a case where it's not an instance of `SpeechSynthesisUtterance`? instanceof is a relatively expensive operation

src/announcer/speechSynthesis.js
```diff
@@ -19,33 +19,69 @@ import { Log } from '../lib/log.js'
...
+    clear(id)
+    return
+  }
+
+  // Clear existing timer for this specific utterance
+  if (state?.timer !== null) {
```
michielvandergeest: Remove the optional chaining syntax

src/announcer/speechSynthesis.js
```diff
@@ -57,45 +93,124 @@ const defaultUtteranceProps = {
 }
 
 const initialize = () => {
+  // syn api check: syn might not have getVoices method
+  if (!syn || typeof syn.getVoices !== 'function') {
+    initialized = true
```
michielvandergeest: is this correct? if `syn` is falsy, do we want to set `initialized` to `true`?

src/announcer/speechSynthesis.js
```diff
@@ -57,45 +93,124 @@ const defaultUtteranceProps = {
...
+    Log.warn('SpeechSynthesis - waiting for ready state...')
+
+    const startTime = Date.now()
+
+    const intervalId = window.setInterval(() => {
+      const elapsed = Date.now() - startTime
+      const isReady = !syn.speaking && !syn.pending
+
+      if (isReady) {
+        Log.warn(`SpeechSynthesis - ready after ${elapsed}ms`)
```
michielvandergeest: These should probably be `Log.debug` statements and not `Log.warn`?

src/announcer/speechSynthesis.js
```diff
@@ -57,45 +93,124 @@ const defaultUtteranceProps = {
...
+      Log.warn('SpeechSynthesis - ready immediately')
+      resolve()
+      return
+    }
+
+    Log.warn('SpeechSynthesis - waiting for ready state...')
+
+    const startTime = Date.now()
+
+    const intervalId = window.setInterval(() => {
```
michielvandergeest: why use `window.setInterval` instead of `setInterval`?

src/announcer/speechSynthesis.js
```diff
@@ -57,45 +93,124 @@ const defaultUtteranceProps = {
...
+      const elapsed = Date.now() - startTime
+      const isReady = !syn.speaking && !syn.pending
+
+      if (isReady) {
+        Log.warn(`SpeechSynthesis - ready after ${elapsed}ms`)
+        window.clearInterval(intervalId)
```
michielvandergeest: why use `window.clearInterval` instead of `clearInterval`?

src/announcer/speechSynthesis.js
```diff
@@ -57,45 +93,124 @@ const defaultUtteranceProps = {
...
-      utterances.delete(id) // Cleanup
-      resolve()
+    utterance.onend = (result) => {
+      clear(id)
+      resolve(result)
     }
 
     utterance.onerror = (e) => {
-      clear()
-      utterances.delete(id) // Cleanup
```
michielvandergeest, line 94: is this not necessary anymore?

src/announcer/announcer.js
tuncayuk: Yes, the timeout was increased from 200ms to 300ms to provide better "easing when speaking the messages to reduce stuttering" (as noted in the comment on announcer.js).

As you know the debounce mechanism prevents rapid-fire speech synthesis calls when messages are queued in quick succession. 

After this change;

1. Gives the speech synthesis engine more time to stabilize between utterances

2. Still provides quick enough feedback for users (only 100ms longer)

3. Different platforms/browsers have varying speech synthesis initialization times

The original 200ms was functional but could still exhibit stuttering behavior in certain scenarios, particularly when multiple announcements were queued rapidly. The 300ms value provides a better balance between smoothness and responsiveness.

**During my tests, I predicted that a response time of 300ms would result in more stable results. This change is not mandatory, just an improvement for avoiding different scenarios**

src/announcer/announcer.js
tuncayuk: Yes, But I have added cancelPrevious param to speak API. Default values is false, so it's not breaking change.

Related commit: https://github.com/lightning-js/blits/pull/544/commits/6e128c384e272c2831f0cb4b0c8c59181aada3f9

src/announcer/speechSynthesis.js
tuncayuk: Done

src/announcer/speechSynthesis.js
tuncayuk: Done

src/announcer/speechSynthesis.js
tuncayuk: I have removed below part, it's not mandatory. Thanks

```
  if (!(utterance instanceof SpeechSynthesisUtterance)) {
    clear(id)
    return
  }
```

src/announcer/speechSynthesis.js
tuncayuk: Good catch. I have changed to `initialized = false`. Thanks!

src/announcer/speechSynthesis.js
tuncayuk: Done

src/announcer/speechSynthesis.js
tuncayuk: I guess, we might keep `window.setInterval`  and `window.clearInterval` here. 

The TypeScript error occurs because:

Plain `setInterval`  can have Node.js typings which return NodeJS.Timer.

`window.setInterval`  explicitly uses browser typings which return number.

Note: When I remove the window, getting the below warning.

Screenshot 2026-01-02 at 15 03 50

src/announcer/speechSynthesis.js
tuncayuk: Same as above.

src/announcer/speechSynthesis.js
tuncayuk: New implementation calls clear API which is deletes utterance and related timer which is correct.

```
    utterance.onerror = (e) => {
      Log.warn('SpeechSynthesisUtterance error:', e)
      clear(id)
      resolve()
    }
```

clear API
```
const clear = (id) => {
  const state = utterances.get(id)
  if (!state) {
    return
  }
  if (state && state.timer !== null) {
    clearTimeout(state.timer)
    state.timer = null
  }
  utterances.delete(id)
}
```

src/announcer/announcer.js
```diff
@@ -26,7 +26,9 @@ let currentId = null
 let debounce = null
 
 // Global default utterance options
-let globalDefaultOptions = {}
+let globalDefaultOptions = {
+  enableUtteranceKeepAlive: /android/i.test((window.navigator || {}).userAgent || ''),
```
tuncayuk: I guess you made below change but default enableUtteranceKeepAlive value should not contain `Android` instead `contain android`.

Screenshot 2026-01-02 at 15 10 19

tuncayuk: default value was like below.

```
const defaultUtteranceKeepAlive = !isAndroid; 
```

src/announcer/announcer.js
michielvandergeest: Ok! Thanks for the explanation. That makes sense

src/announcer/announcer.js
michielvandergeest: Yes correct, I missed the `!` in my update. Can you change it back, please?

tuncayuk: Revoked. Thanks!

PR #545, opened 2025-11-27
Addressed the issue with using a computed prop as the source for loop data.

suresh-gangumalla:
Observed that while using computed property as for-loop data collection then for-loop effect is not tracking for all the other instance variables which are using in computed property function defintion. Ideally changes to any property inside computed function should execute for-loop again, This PR will actually collect list of those component refferences inside each computed method and track for-loop for all those component variables.


HadiAlMarzooq:
Hi @suresh-gangumalla, I've already addressed this in https://github.com/lightning-js/blits/pull/541, can you check and verify?

PR #547, opened 2025-11-28
added reactivity support to regular shaders

jfboeve:
Added reactivity for shader attribute.
This only works for stand alone shaders, not effects (dynamic shader)

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -383,7 +383,7 @@ const propsTransformer = {
     const target = this.element.node !== undefined ? this.element.node : this.props
 
     if (v === null) {
-      target['shader'] = renderer.createShader('DefaultShader')
+      target['shader'] = null
       return
     }
 


src/engines/L3/element.js
```diff
@@ -380,10 +380,21 @@ const propsTransformer = {
...
-    if (v !== null) {
-      this.props['shader'] = renderer.createShader(v.type, v.props)
-    } else {
-      this.props['shader'] = renderer.createShader('DefaultShader')
+    const target = this.element.node !== undefined ? this.element.node : this.props
+
+    if (v === null) {
+      target['shader'] = renderer.createShader('DefaultShader')
+      return
+    }
```
michielvandergeest, line 388: do we need to create the default shader when the value of shader is `null`? isn't that the default already?

also, is it possible to first have a `null` value and then define the shader afterwards?

jfboeve: You are correct. Yeah it's always possible to change shaders during runtime

PR #550, opened 2025-12-02
feat(router): add afterEach hook to RouterHooks for post-navigation handling

tuncayuk:
**Summary**

Adds a new afterEach hook to the router that executes after navigation completes, providing access to both the route objects and their corresponding component instances.

**Changes**

- Added afterEach and after hook execution in the navigate function
- The hook is called after the new view is set as the active view but before focus is transferred
- Hook receives an object with:
  - to: The destination route
  - toComponent: The destination component instance
  - from: The previous route (or undefined if navigating to the first route)
  - fromComponent: The previous component instance (or null if reusing or no previous view)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -626,7 +626,7 @@ declare module '@lightningjs/blits' {
   export interface RouterHooks {
     init?: () => Promise<void> | void;
     beforeEach?: (to: Route, from: Route) => string | Route | Promise<string | Route> | void;
-    afterEach?: (to: Route, toComponent: ComponentBase, from: Route | undefined, fromComponent: ComponentBase | null) => string | Route | Promise<string | Route> | void;
+    afterEach?: (to: Route, from: Route) => string | Route | Promise<string | Route> | void;
     error?: (err: string) => string | Route | Promise<string | Route> | void;
   }
 
@@ -750,7 +750,7 @@ declare module '@lightningjs/blits' {
 
   export interface RouteHooks {
     before?: (to: Route, from: Route) => string | Route | Promise<string | Route>;
-    after?: (to: Route, toComponent: ComponentBase, from: Route | undefined, fromComponent: ComponentBase | null) => string | Route | Promise<string | Route>;
+    after?: (to: Route, from: Route) => string | Route | Promise<string | Route>;
   }
 
   export type Route = {
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -511,9 +511,7 @@ export const navigate = async function () {
             await hooks.afterEach.call(
               this.parent,
               route, // to
-              view, // toComponent
-              previousRoute, // from
-              oldView // fromComponent
+              previousRoute // from
             )
           } catch (error) {
             Log.error('Error in "AfterEach" Hook', error)
@@ -526,9 +524,7 @@ export const navigate = async function () {
           await route.hooks.after.call(
             this.parent,
             route, // to
-            view, // toComponent
-            previousRoute, // from
-            oldView // fromComponent
+            previousRoute // from
           )
         } catch (error) {
           Log.error('Error or Rejected Promise in "After" Hook', error)


src/router/router.js
```diff
@@ -459,6 +460,41 @@ export const navigate = async function () {
       const children = this[symbols.children]
       this.activeView = children[children.length - 1]
 
+      if (this.parent[symbols.routerHooks]) {
+        const hooks = this.parent[symbols.routerHooks]
+        if (hooks.afterEach) {
+          try {
+            // Get the previous view before it's removed
+            const previousView =
```
il-sairamg: On first navigation, previousView may incorrectly resolve to the root container since children includes it at index 0.
Consider gating on previousRoute or using a stricter check (e.g. children.length > 2) so previousView is null when no previous route exists — the same applies to the route.hooks.after block below.

src/router/router.js
```diff
@@ -459,6 +460,41 @@ export const navigate = async function () {
...
+            toComponent: view,
+            from: previousRoute,
+            fromComponent: previousView,
+          })
+        } catch (error) {
+          Log.error('Error or Rejected Promise in "After" Hook', error)
+        }
+      }
+
       // set focus to the view that we're routing to (unless explicitly disabling passing focus)
```
il-sairamg, line 463: “after” hooks currently run before focus is applied and transitions complete, so they may observe an intermediate UI state. Consider running them after focus/transition work finishes, or document that they run after view creation rather than after navigation completion.

src/router/router.js
tuncayuk: Added a check for `previousRoute` to ensure we only get the `previousView` when there's actually a previous route. 

Thanks!

src/router/router.js
tuncayuk: Moved afterEach and after hooks to the end of the transition.

Thanks!

src/router/router.js
```diff
@@ -500,6 +501,45 @@ export const navigate = async function () {
           await setOrAnimate(holder, route.transition.in, shouldAnimate)
         }
       }
+
+      if (this.parent[symbols.routerHooks]) {
+        const hooks = this.parent[symbols.routerHooks]
```
il-sairamg, line 508: removeView() isn’t awaited, so “after” hooks can run while the exit transition / cleanup is still in progress. Consider awaiting cleanup.

src/router/router.js
```diff
@@ -500,6 +501,45 @@ export const navigate = async function () {
...
+        if (hooks.afterEach) {
+          try {
+            // Get the previous view before it's removed
+            const previousView =
```
il-sairamg: previousView is derived from children after mutation, which can be fragile and may resolve to the root container. Capturing the removed oldView would be more reliable. Same applies to the route.hooks.after block below.

index.d.ts
```diff
@@ -626,6 +626,7 @@ declare module '@lightningjs/blits' {
   export interface RouterHooks {
     init?: () => Promise<> | void;
     beforeEach?: (to: Route, from: Route) => string | Route | Promise<string | Route> | void;
+    afterEach?: (to: Route, toComponent: ComponentBase, from: Route, fromComponent: ComponentBase) => string | Route | Promise<string | Route> | void;
```
il-sairamg: Typing: from / fromComponent can be undefined / null at runtime (first nav, reuse). Consider changing both afterEach and after to from?: Route and fromComponent: ComponentBase | null (or guarantee non-null in implementation).

src/router/router.js
tuncayuk: Used `oldView` instead of `previousView`. Thanks.

index.d.ts
tuncayuk: Type definition changed as requested. Thanks!

src/router/router.js
tuncayuk: Changed `removeView()` to await.

index.d.ts
```diff
@@ -626,6 +626,7 @@ declare module '@lightningjs/blits' {
   export interface RouterHooks {
     init?: () => Promise<> | void;
```
il-sairamg: Promise<> is invalid TypeScript syntax; use Promise<void> instead.

src/router/router.js
```diff
@@ -500,6 +503,37 @@ export const navigate = async function () {
           await setOrAnimate(holder, route.transition.in, shouldAnimate)
         }
       }
+
```
il-sairamg, line 506: Line 498: route.transition.length is undefined because route.transition is an object.
Using transition.in.length - 1 ensures the last in-transition is awaited.

index.d.ts
```diff
@@ -749,6 +750,7 @@ declare module '@lightningjs/blits' {
 
   export interface RouteHooks {
     before?: (to: Route, from: Route) => string | Route | Promise<string | Route>;
+    after?: (to: Route, toComponent: ComponentBase, from: Route | undefined, fromComponent: ComponentBase | null) => string | Route | Promise<string | Route>;
```
michielvandergeest: Why is the signature of the `after` hook not the same as the `before` hook, with just `to` and `from`? That would make more sense to me and make the API more predictable

index.d.ts
tuncayuk: Done

src/router/router.js
tuncayuk: Done

index.d.ts
tuncayuk: My goal is to use the after navigation hook to trigger shared, post-navigation API calls from the destination component (toComponent).

For example, if a component implements an analytics API like sendAnalytics(), we can automatically call it after the route change completes—so each page can own its own analytics logic, while the hook handles the common wiring.

**Example:**
```ts
after(to, toComponent, from, fromComponent) {
  // Send analytics after navigating to the new page,
  // but only if the destination component exposes `sendAnalytics`.
  toComponent?.sendAnalytics?.();
}
```

If you want I can remove new params.

michielvandergeest: Sorry for the late reply @tuncayuk 

I think I prefer removing the params for now and keeping the signature the same. If the use case described arrises I'd prefer thinking of a different approach rather than making the signatures different.

Thanks!

PR #551, opened 2025-12-02
test(sprite): add tests for texture reuse, mapping, and missing frames

il-sairamg:
Adds tests covering Sprite texture creation, reuse, and SubTexture mapping.
Includes cases for missing frames, invalid inputs, and fallback behavior.

### DIFF
--- a/src/components/Sprite.js
+++ b/src/components/Sprite.js
@@ -66,10 +66,12 @@ export default () =>
           this.frame !== null &&
           this.frame !== undefined
         ) {
-          if ('frames' in this.map && this.map.frames !== null && this.map.frames !== undefined) {
-            if (this.frame in this.map.frames) {
-              options = Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
-            }
+          if (
+            this.map.frames !== null &&
+            this.map.frames !== undefined &&
+            this.frame in this.map.frames
+          ) {
+            options = Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
           } else if (this.frame in this.map) {
             options = this.map[this.frame]
           }


src/components/Sprite.js
```diff
@@ -61,15 +61,18 @@ export default () =>
...
+          this.map !== undefined &&
+          this.frame !== null &&
+          this.frame !== undefined
         ) {
-          options =
-            'frames' in this.map
-              ? Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
-              : this.map[this.frame]
+          if ('frames' in this.map && this.map.frames !== null && this.map.frames !== undefined) {
+            if (this.frame in this.map.frames) {
```
michielvandergeest: is it necessary for this condition to be nested? or can it be included in the condition one level up?

il-sairamg: Reduced nesting by lifting common guards; preserved existing behavior.

src/components/Sprite.js
```diff
@@ -61,15 +61,20 @@ export default () =>
...
-              : this.map[this.frame]
+          if (
+            this.map.frames !== null &&
+            this.map.frames !== undefined &&
+            this.frame in this.map.frames
+          ) {
+            options = Object.assign({}, this.map.defaults || {}, this.map.frames[this.frame])
+          } else if (this.frame in this.map) {
+            options = this.map[this.frame]
+          }
```
michielvandergeest, line 77: can you explain these changes? I'm not sure I follow the changes (and what they fix, that was previously broken) 🙂

src/components/Sprite.js
```diff
@@ -61,15 +61,20 @@ export default () =>
         // Resolve frame data from sprite map
         let options = null
         if (
-          this.map !== undefined &&
           this.map !== null &&
-          this.frame !== undefined &&
-          this.frame !== null
+          this.map !== undefined &&
+          this.frame !== null &&
+          this.frame !== undefined
```
michielvandergeest, line 67: here we just changed the order of the conditions in the if, right?

src/components/Sprite.js
il-sairamg: Hi @michielvandergeest, Earlier, we only checked if frames existed on map, which could still be null and cause a runtime error when accessed.
The new code ensures map.frames actually exists and contains the frame before using it, preventing crashes and making frame resolution safer.

src/components/Sprite.js
il-sairamg: Yes, this is just a reordering of the conditions.

PR #554, opened 2025-12-16
feat(router): add runtime navigateHistory toggle and forward Back to …

il-sairamg:
Adds a runtime router property ($router.backNavigation) to control RouterView's automatic Back-history navigation (default remains enabled).
Updates RouterView to forward Back events to the parent when history navigation is disabled or unavailable, allowing app-level handling.

### DIFF
--- a/src/components/RouterView.js
+++ b/src/components/RouterView.js
@@ -57,12 +57,12 @@ export default () =>
       input: {
         back(e) {
           if (routerState.backNavigation === false) {
-            this.parent?.$input?.(e)
+            this.parent.$input(e)
             return
           }
           const navigating = Router.back.call(this)
           if (navigating === false) {
-            this.parent?.$focus?.(e)
+            this.parent.$focus(e)
           }
         },
       },
--- a/docs/router/hooks.md
+++ b/docs/router/hooks.md
@@ -77,6 +77,29 @@ In order to configure global router hooks the `routes` key in the Application co
 
 Alongside the `router.routes` key we can now define a `router.hooks` object, which can have any of the following pre-defined hook functions:
 
+### Router Settings
+
+The router configuration also supports router settings alongside `router.routes` and `router.hooks`:
+
+- `backNavigation` - Enable or disable RouterView history navigation on Back input (default: `true`). When set to `false`, the Back input will be passed to the parent component instead of navigating back through the history stack. This setting can be configured in the router config or changed at runtime via `this.$router.backNavigation`.
+
+> **Note:** `backNavigation` is an app-wide setting that affects all `RouterView` instances in your application, as the router state is global and shared.
+
+```js
+export default Blits.Application({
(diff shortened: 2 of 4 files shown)


src/components/RouterView.js
```diff
@@ -16,46 +16,54 @@
...
+      input: {
+        back(e) {
+          if (routerState.navigateHistory !== false && Router.back.call(this)) return
+
+          // If history navigation is disabled or fails, forward Back to parent
+          /** @type {RouterViewParent | undefined | null} */
+          const parent = this.parent
+          const handledByParent = parent?.$input?.(e) ?? false
+          if (handledByParent === false) parent?.$focus?.(e)
+        },
```
michielvandergeest, line 67: I'm not sure I understand these changes. I don't think the parent should get focus when the key press isn't handled by the parent?

Shouldn't we just do this instead:

```js
back(e) {
          if (routerState.navigateHistory === false) {
            this.parent.$input(e)
            return
          }
          const navigating = Router.back.call(this)
          if (navigating === false) {
            this.parent.$focus(e)
          }
}
```

src/component/base/router.js
```diff
@@ -16,13 +16,19 @@
...
+import { to, currentRoute, back, state, setNavigateHistory } from '../../router/router.js'
 
 export default {
   $router: {
     value: {
       to,
       back,
+      navigateHistory(enabled = true) {
+        setNavigateHistory(enabled)
+      },
```
michielvandergeest: can we make this a `setter` instead of a function? also I'm thinking of naming it `backNavigation` to be shorter and more intutive

that way we'll be able to use it as `this.$router.backNavigation = true / false` and `if(this.$router.backNavigation === true)`

src/components/RouterView.js
il-sairamg: Good point — my earlier approach could move focus even when the parent didn’t handle Back. Your version is clearer and handles input and focus separately.

src/component/base/router.js
il-sairamg: Updated as suggested. Thanks!

src/components/RouterView.js
```diff
@@ -16,46 +16,55 @@
...
-    input: {
-      back(e) {
-        const navigating = Router.back.call(this)
-        if (navigating === false) {
-          this.parent.$focus(e)
-        }
+      input: {
+        back(e) {
+          if (routerState.backNavigation === false) {
+            this.parent?.$input?.(e)
```
michielvandergeest: I don't want us to use the optional chaining syntax, due to the inefficient way it's tranpiled into ES5 code

Also, is there a reasonable scenario where parent would be undefined?

il-sairamg: Agreed. RouterView always has a parent, so optional chaining isn’t needed.
The checks were only for runtime safety.

PR #555, opened 2025-12-17
feat(types): add ComponentCustomProperties augmentation point and typ…

il-sairamg:
Added a public TypeScript extension point and exported built-in plugin types for safe consumer augmentation.

Typed @lightningjs/blits/plugins and cleaned up .d.ts for correctness and IntelliSense.

### DIFF
--- a/docs/plugins/global_app_state.md
+++ b/docs/plugins/global_app_state.md
@@ -50,7 +50,7 @@ The newly created Global App state works exactly the same as an internal Compone
 
 ## TypeScript Support
 
-Enable autocomplete and type inference for the App State plugin by adding a `blits.d.ts` file:
+Enable autocomplete and type inference for the App State plugin by adding a `blits.d.ts` file in the root folder of your app project:
 
 ```typescript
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -365,12 +365,44 @@ declare module '@lightningjs/blits' {
   // Extension point for app- and plugin-specific fields on the component `this`.
   // Add your own properties (e.g., `$telemetry`, `componentName`) via TypeScript
   // module augmentation in your app, without changing core types.
+  // Public extension point for app/plugin augmentation via module augmentation.
   // Note: `ComponentBase` extends this interface, so augmented fields appear in all
   // hooks, methods, input, computed, and watch.
-  export interface CustomComponentProperties {
+  export interface ComponentCustomProperties {
```
michielvandergeest: Please keep the name `CustomComponentProperties`

michielvandergeest (changes requested):

index.d.ts
il-sairamg: Updated as suggested. Thanks!

PR #563, opened 2025-12-31
Added test cases for prepublishOnly script

suresh-gangumalla:
1. resovled issue with including .test.js files inside node-modules directory in windows OS
2. Exporting pre publish functions to add unit test cases
3. Using Sinon Stubs to mimic different function returns instead of depending on original functions implementation which results to reading, copying, writing .org.js files of src/components. With stubs mimic the behaviour to match the scope of test case

### DIFF
--- a/scripts/prepublishOnly.js
+++ b/scripts/prepublishOnly.js
@@ -17,6 +17,7 @@
 
 import fs from 'fs'
 import path from 'path'
+import { fileURLToPath } from 'url'
 import compiler from '../src/lib/precompiler/precompiler.js'
 import { exec } from 'child_process'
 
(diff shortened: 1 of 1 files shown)


scripts/prepublishOnly.js
```diff
@@ -84,4 +84,9 @@ function formatFileWithESLint(filePath) {
   })
 }
 
-precompileComponents()
+// Only run if this file is executed directly (not imported)
+if (import.meta.url === `file:///${process.argv[1].replace(/\\/g, '/')}`) {
+  precompileComponents()
+}
```
michielvandergeest: this check doesn't seem to work during the `npm run release` step, meaning no pre-compilation is happening during release

the replace is appending an extra slash `/` compared to the original `import.meta.url `

Image

PR #564, opened 2026-01-05
Merge dev into v2/update

suresh-gangumalla:


element-test-full.txt
michielvandergeest, line 1: this file should probably be deleted

index.d.ts
```diff
@@ -364,7 +366,39 @@ declare module '@lightningjs/blits' {
...
+    get<T>(key: string, fallback: T): T
+    set(theme: string): void
+  }
+
+  export interface StoragePlugin {
+    get<T = unknown>(key: string): T | null
+    set(key: string, value: unknown): boolean
+    remove(key: string): void
+    clear(): void
+  }
```
michielvandergeest: should these still be here actually? Maybe also they shouldn't exist in the current master version (since we've changed this to explicit imports).

@il-sairamg can you comment?

test-output.txt
michielvandergeest, line 1: remove this file?

index.d.ts
il-sairamg: You're correct — these shouldn't be in index.d.ts anymore.

PR #567, opened 2026-01-07
feat: add  method for component-scoped debouncing

il-sairamg:
Implements key-based debounce API ($debounce(key, fn, ms)) that automatically clears timers on component destroy. 
Includes $clearDebounce(key) and $clearDebounces() methods and documentations.

### DIFF
--- a/docs/components/utility_methods.md
+++ b/docs/components/utility_methods.md
@@ -446,7 +446,7 @@ The method returns a `debounce id`, which can be used to manually clear the debo
 **Key Features:**
 - **Name-based**: Each debounce is identified by a unique name (unique per component instance)
 - **Automatic replacement**: Calling `$debounce` with the same name replaces the previous debounce
-- **Memory efficient**: Only stores timeout IDs, function is captured in closure
+- **Memory efficient**: Only stores debounce IDs internally, function is captured in closure
 - **Automatic cleanup**: All debounces are cleared when component is destroyed
 
 ### $clearDebounce


src/component/base/timeouts_intervals.js
```diff
@@ -96,4 +96,59 @@ export default {
...
     configurable: false,
   },
+  $debounce: {
+    value: function (key, fn, ms, ...params) {
+      // early exit when component is marked as end of life
+      if (this.eol === true) return
+
+      // clear existing debounce for this key if it exists
+      const existing = this[symbols.debounces].get(key)
+      if (existing) {
```
michielvandergeest: we can make an explicit not `undefined` check here

src/component/base/timeouts_intervals.js
```diff
@@ -96,4 +96,59 @@ export default {
...
+      return timeoutId
+    },
+    writable: false,
+    enumerable: true,
+    configurable: false,
+  },
+  $clearDebounce: {
+    value: function (key) {
+      const existing = this[symbols.debounces].get(key)
+      if (existing) {
```
michielvandergeest: same comment about the not `undefined` check

src/component/base/timeouts_intervals.js
```diff
@@ -96,4 +96,59 @@ export default {
...
+    writable: false,
+    enumerable: true,
+    configurable: false,
+  },
+  $clearDebounces: {
+    value: function () {
+      // clear all timeouts associated with debounces
+      for (const timeoutId of this[symbols.debounces].values()) {
+        this.$clearTimeout(timeoutId)
+      }
```
michielvandergeest: `for of` is not the most optimal way. I think it's better to collect the values in a variable and and do a regular `for` or `while` loop

index.d.ts
```diff
@@ -427,6 +427,26 @@ declare module '@lightningjs/blits' {
...
+    $debounce: (key: string, callback: (...args: any[]) => void, ms?: number, ...args: any[]) => ReturnType<typeof setTimeout>
+
+    /**
+    * Clear a specific debounce by key
+    * @param key - The key of the debounce to clear
+    */
+    $clearDebounce: (key: string) => void
+
+    /**
+    * Clear all debounces (automatically called on component destroy)
```
michielvandergeest: Clear all debounces registered _on the component_ (right?)

docs/components/utility_methods.md
```diff
@@ -428,3 +428,73 @@ export default Blits.Component('MyComponent', {
...
+
+Similar to timeouts and intervals, debounced functions can cause memory leaks if not properly cleaned up. Blits provides built-in debounce methods that automatically handle cleanup when a component is destroyed.
+
+### $debounce
+
+The `this.$debounce()`-method creates a debounced function that delays execution until after a specified delay has passed since the last invocation. If the same key is debounced again before the delay completes, the previous debounce is cancelled and a new one is created.
+
+The first argument is a `key` (string) that uniquely identifies this debounce instance. The second argument is the `callback` function to execute. The third argument is the `delay` in milliseconds. Additional arguments can be passed and will be forwarded to the callback function.
+
+The method returns a `timeout id`, which can be used to manually clear the debounce.
```
michielvandergeest: it's technically a timeout id, but maybe we can call it a `debounce id`?

docs/components/utility_methods.md
```diff
@@ -428,3 +428,73 @@ export default Blits.Component('MyComponent', {
...
+
+Debouncing is a technique to limit the rate at which a function executes. This is particularly useful when navigating through lists or handling rapid user input, where you want to execute a method only after the user has stopped the action for a specified delay.
+
+Similar to timeouts and intervals, debounced functions can cause memory leaks if not properly cleaned up. Blits provides built-in debounce methods that automatically handle cleanup when a component is destroyed.
+
+### $debounce
+
+The `this.$debounce()`-method creates a debounced function that delays execution until after a specified delay has passed since the last invocation. If the same key is debounced again before the delay completes, the previous debounce is cancelled and a new one is created.
+
+The first argument is a `key` (string) that uniquely identifies this debounce instance. The second argument is the `callback` function to execute. The third argument is the `delay` in milliseconds. Additional arguments can be passed and will be forwarded to the callback function.
```
michielvandergeest: Would `name` be more intuitive than `key`, maybe? Not sure tbh

Also, the name or key should be unique per component instance, correct?

docs/components/utility_methods.md
```diff
@@ -428,3 +428,73 @@ export default Blits.Component('MyComponent', {
...
+    }
+  },
+  hooks: {
+    unfocus() {
+      // Optionally clear all debounces when component loses focus
+      this.$clearDebounces()
+    }
+  }
+})
+```
```
michielvandergeest, line 500: Nice example! 👏

index.d.ts
il-sairamg: yes. It clears all debounces registered on the component instance.

docs/components/utility_methods.md
il-sairamg: Yes, exactly. The name is unique per component instance—each component has its own debounces Map.

src/component/base/timeouts_intervals.js
il-sairamg: Updated as suggested, Thanks!

docs/components/utility_methods.md
il-sairamg: Updated as suggested.

michielvandergeest (approved):
LGTM!

PR #569, opened 2026-01-13
feat: add automatic inspector metadata keys

il-sairamg:
Automatically set inspector metadata for all elements:
- $componentType: element/component type
- $hasFocus: component focus state (auto-updates)
- $isTransitioning: element transition state (auto-updates)

Only active in dev mode when inspector is enabled.

### DIFF
--- a/src/lib/lifecycle.js
+++ b/src/lib/lifecycle.js
@@ -18,6 +18,9 @@
 import { Log } from './log.js'
 import { emit, privateEmit } from './hooks.js'
 import symbols from './symbols.js'
+import Settings from '../settings.js'
+
+let inspectorEnabled = null
 
 /**
  * List of valid lifecycle states for a component.
@@ -82,10 +85,15 @@ export default {
       // emit 'public' hook
       emit(v, this.component[symbols.identifier], this.component)
       // update the built-in hasFocus state variable
+      if (v === 'focus' || v === 'unfocus') {
+        if (inspectorEnabled === null) {
+          inspectorEnabled = Settings.get('inspector', false)
+        }
+      }
       if (v === 'focus') {
         this.component[symbols.state].hasFocus = true
-        // Update inspector metadata on component's holder element
         if (
+          inspectorEnabled === true &&
           this.component[symbols.holder] &&
           typeof this.component[symbols.holder].setInspectorMetadata === 'function'
         ) {
@@ -94,8 +102,8 @@ export default {
       }
       if (v === 'unfocus') {
         this.component[symbols.state].hasFocus = false
-        // Update inspector metadata on component's holder element
         if (
+          inspectorEnabled === true &&
           this.component[symbols.holder] &&
           typeof this.component[symbols.holder].setInspectorMetadata === 'function'
         ) {


src/lib/codegenerator/generator.js
```diff
@@ -170,6 +170,18 @@ const generateElementCode = function (
     renderCode.push(`elementConfigs[${counter}][Symbol.for('isSlot')] = true`)
   }
 
+  // Generate framework inspector data setup (dev mode only)
+  if (isDev === true) {
+    const componentType = templateObject[Symbol.for('componentType')] || 'Element'
```
michielvandergeest: Maybe we should set this for Elements, but only components. That doesn't add too much value and should make this a bit more light-weight

src/lib/codegenerator/generator.js
```diff
@@ -170,6 +170,18 @@ const generateElementCode = function (
...
+    renderCode.push(`
+      if (${elm} !== undefined && typeof ${elm}.setInspectorMetadata === 'function') {
```
michielvandergeest: is the `typeof ${elm}.setInspectorMetadata === 'function'` check really necessary?

src/lib/lifecycle.js
```diff
@@ -82,8 +82,26 @@ export default {
...
       emit(v, this.component[symbols.identifier], this.component)
       // update the built-in hasFocus state variable
-      if (v === 'focus') this.component[symbols.state].hasFocus = true
-      if (v === 'unfocus') this.component[symbols.state].hasFocus = false
+      if (v === 'focus') {
+        this.component[symbols.state].hasFocus = true
+        // Update inspector metadata on component's holder element
+        if (
+          this.component[symbols.holder] &&
+          typeof this.component[symbols.holder].setInspectorMetadata === 'function'
```
michielvandergeest, line 98: do we need the typeof function check here?

src/engines/L3/element.js
```diff
@@ -564,6 +564,44 @@ const Element = {
...
+  /**
+   * Sets framework-provided inspector metadata
+   * Only sets if inspector is enabled and in dev mode
+   * @param {Object} data - Framework inspector metadata to merge
+   */
+  setInspectorMetadata(data) {
+    // Early return if inspector not enabled (performance optimization)
+    if (inspectorEnabled === false) {
+      return
+    }
```
michielvandergeest: double check? since we don't call this function is inspectorEnabled is false to begin with?

src/engines/L3/element.js
```diff
@@ -564,6 +564,44 @@ const Element = {
...
+
+    if (inspectorEnabled === null) {
+      inspectorEnabled = Settings.get('inspector', false)
+      if (inspectorEnabled === false) return
+    }
```
michielvandergeest: is this a condition that can happen?

il-sairamg: I’ve updated this accordingly so the inspectorEnabled === null case is no longer needed here.

src/lib/codegenerator/generator.js
il-sairamg: I've updated this to only set metadata for components (not elements), which makes it lighter-weight

src/lib/codegenerator/generator.js
il-sairamg: Yes — it’s a cheap defensive guard to avoid dev-time crashes in edge cases (tests/mocks/destroyed elements).

src/engines/L3/element.js
il-sairamg: Yes — it’s a safe central guard, so callers can call setInspectorMetadata() without checking inspectorEnabled first.

src/lib/lifecycle.js
```diff
@@ -82,8 +82,26 @@ export default {
       // emit 'public' hook
       emit(v, this.component[symbols.identifier], this.component)
       // update the built-in hasFocus state variable
-      if (v === 'focus') this.component[symbols.state].hasFocus = true
-      if (v === 'unfocus') this.component[symbols.state].hasFocus = false
+      if (v === 'focus') {
+        this.component[symbols.state].hasFocus = true
+        // Update inspector metadata on component's holder element
+        if (
```
michielvandergeest, line 95: should we do an early exit looking for dev mode? To make sure this is not always executed also in production?

il-sairamg: Updated checks for early exit.

PR #572, opened 2026-01-20
Add platform setting to launch configuration

E-Justin:
Blits apps need to provide platform to the renderer

### DIFF
--- a/src/engines/L3/launch.js
+++ b/src/engines/L3/launch.js
@@ -100,7 +100,7 @@ export default (App, target, settings = {}) => {
         textureMemory: textureMemorySettings(settings),
         createImageBitmapSupport: 'auto',
         targetFPS: 'maxFPS' in settings ? settings.maxFPS : 0,
-        platform: settings.platform,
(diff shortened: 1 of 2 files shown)


src/engines/L3/launch.js
```diff
@@ -100,6 +100,7 @@ export default (App, target, settings = {}) => {
         textureMemory: textureMemorySettings(settings),
         createImageBitmapSupport: 'auto',
         targetFPS: 'maxFPS' in settings ? settings.maxFPS : 0,
+        platform: settings.platform,
```
michielvandergeest: do we need a default value? or will undefined (in case not platform is provided in settings) be handled gracefully?

estobbart: it's slightly confusing.. maybe we need a renderer change.. the type is defined as..
```
platform: typeof Platform | null;
```
but it looks to default as WebPlatform..
https://github.com/lightning-js/renderer/blob/main/src/main-api/Renderer.ts#L553
```
platform: settings.platform || WebPlatform,
```

michielvandergeest (approved):

PR #585, opened 2026-01-28
Added test case for debounce feature and updated generator tests

il-sairamg:
Added test cases for Debounce feature.
Updated test case for generator


michielvandergeest (approved):
LGTM!

PR #588, opened 2026-01-29
fix(router): set params to null when route has no params

il-sairamg:
Setting state.params = null correctly resets params and prevents stale watcher data.

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -403,8 +403,7 @@ export const navigate = async function () {
       // Update router state after announcements and final route resolution,
       // right before initializing or restoring the view
       state.path = route.path
-      const nextParams = route.params || {}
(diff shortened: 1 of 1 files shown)


src/router/router.js
```diff
@@ -403,7 +403,8 @@ export const navigate = async function () {
       // Update router state after announcements and final route resolution,
       // right before initializing or restoring the view
       state.path = route.path
-      state.params = route.params || {}
+      const nextParams = route.params || {}
+      state.params = Object.keys(nextParams).length === 0 ? null : nextParams
```
michielvandergeest: can't we do this in one line?

something like `state.params = route.params || null`

or `state.params = Object.keys(route.params).length === 0 ? null route.params`

michielvandergeest (approved):
LGTM!

PR #589, opened 2026-01-30
Debouncing Animate function to avoid frequent invocations

suresh-gangumalla:
changes to any property in transition object like delay, duration, vaule, ease are subject to invocation of animate function. when morethan single property get updated then transition subjected to creation of multiple animation objects which are neccessary so added debounce with configurable timeout so that it can reduce number of animation function invocations.

user can configure animateDebounceDelay  in settings to alter delay timeout in ms, the default value is 50ms

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -229,12 +229,6 @@ let textDefaults = null
 let inspectorEnabled = null
 
 /**
- * Debounce delay for animate function (initialized on first use).
- * @type {number|null}
- */
-let animateDebounceDelay = null
-
-/**
  * @typedef {import('../../component.js').BlitsElement} BlitsElement
  *
  * This is always a BlitsElement
@@ -672,11 +666,7 @@ const Element = {
     this.debounceTimeouts[prop] = setTimeout(() => {
       delete this.debounceTimeouts[prop]
       this._executeAnimation(prop, value, transition)
-    }, animateDebounceDelay)
-
-    Log.debug(
-      `Debounced animation for property "${prop}" - will execute in ${animateDebounceDelay}ms`
-    )
+    }, 0)
   },
   _executeAnimation(prop, value, transition) {
     // check if a transition is already scheduled to run on the same prop
@@ -786,7 +776,7 @@ const Element = {
     for (let i = 0; i < debounceProps.length; i++) {
       clearTimeout(this.debounceTimeouts[debounceProps[i]])
     }
-    this.debounceTimeouts = {}
+    this.debounceTimeouts = null
 
     // Clearing transition end callback functions
     const transitionProps = Object.keys(this.scheduledTransitions)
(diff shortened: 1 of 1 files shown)


src/engines/L3/element.js
```diff
@@ -656,13 +662,35 @@ const Element = {
...
+    // Debounce the animation execution
+    this.debounceTimeouts[prop] = setTimeout(() => {
+      delete this.debounceTimeouts[prop]
+      this._executeAnimation(prop, value, transition)
+    }, animateDebounceDelay)
+
+    Log.debug(
+      `Debounced animation for property "${prop}" - will execute in ${animateDebounceDelay}ms`
+    )
+  },
```
michielvandergeest: I think we can leave out this log

src/engines/L3/element.js
```diff
@@ -228,6 +228,12 @@ let textDefaults = null
  */
 let inspectorEnabled = null
 
+/**
+ * Debounce delay for animate function (initialized on first use).
+ * @type {number|null}
+ */
+let animateDebounceDelay = null
```
michielvandergeest: don't need this I think?

src/engines/L3/element.js
```diff
@@ -752,6 +780,14 @@ const Element = {
     if (this.node === null) return
 
     Log.debug('Deleting Node', this.nodeId)
+
+    // Clear all pending debounce timeouts
+    const debounceProps = Object.keys(this.debounceTimeouts)
+    for (let i = 0; i < debounceProps.length; i++) {
+      clearTimeout(this.debounceTimeouts[debounceProps[i]])
+    }
+    this.debounceTimeouts = {}
```
michielvandergeest: maybe `null` it upon cleanup

src/engines/L3/element.js
```diff
@@ -831,9 +867,13 @@ export default (config, component) => {
   if (inspectorEnabled === null) {
     inspectorEnabled = Settings.get('inspector', false)
   }
+  if (animateDebounceDelay === null) {
+    animateDebounceDelay = Settings.get('animateDebounceDelay', 0)
+  }
```
michielvandergeest: don't think this has to be a setting

michielvandergeest (changes requested):

src/engines/L3/element.js
suresh-gangumalla: Done

src/engines/L3/element.js
suresh-gangumalla: Yes, as we are always going with 0 ms timeout we don't need this. Removed in latest commit.

src/engines/L3/element.js
suresh-gangumalla: Sure

src/engines/L3/element.js
suresh-gangumalla: Yes, remove this statement.

PR #594, opened 2026-02-05
Support expression-based @event handlers alongside method references

il-sairamg:
- Enable expression-based event handlers for @ listeners while preserving direct method binding behavior.


michielvandergeest (approved):
LGTM!

PR #645, opened 2026-03-23
reactivity guard: adds optional chaining for injected variables 

uguraslan:
Computed properties with null/undefined guards were crashing at runtime because
the injected guard expression (e.g. `this.foo.bar;`) runs before the
user's own null checks and throws error when intermediate values are absent.

Guards are now injected with optional chaining: `this.foo?.bar;`


uguraslan:
We decided not to go on with optional chaining for this fix.

PR #647, opened 2026-03-23
reactivity guard: adds `&&` chaining for injected variables

uguraslan:
Computed properties with null/undefined guards were crashing at runtime because the injected guard expression (e.g. `this.foo.bar;`) runs before the user's own null checks and throws error when intermediate values are absent.

Variables are now injected as a `&&` chain: `this.foo && this.foo.bar;`.


michielvandergeest (approved):
LGTM!

PR #649, opened 2026-03-27
fix wrong props targeted by shadow/border

jfboeve:
Fixed an issue where the wrong shader props were target by shader / border when not combined with rounded shader.


michielvandergeest (approved):
LGTM

PR #650, opened 2026-03-31
Feature/router simplification

michielvandergeest:
WIP PR to simplify the code of the router which has grown a bit out of proportion over time

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -119,7 +119,7 @@ export const navigate = async function () {
   )
 
   // early return when route not found
-  if (route == false) {
+  if (route === false) {
     state.navigating = false
 
     Log.error(`Route ${route.hash} not found`)
@@ -214,11 +214,10 @@ export const navigate = async function () {
   // Announce route change if a message has been specified for this route
   if (route.announce) {
     if (typeof route.announce === 'string') {
-      route.announce = {
-        message: route.announce,
-      }
+      Announcer.speak(route.announce)
+    } else {
+      Announcer.speak(route.announce.message, route.announce.politeness)
     }
-    Announcer.speak(route.announce.message, route.announce.politeness)
   }
 
   // Update router state after announcements and final route resolution,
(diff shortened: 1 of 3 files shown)


src/router/router.js
```diff
@@ -250,316 +99,247 @@ const makeRouteObject = (route, overrides) => {
...
-          preventHashChangeNavigation = true
-          currentRoute = previousRoute
-          window.history.back()
-
-          navigatingBack = false
-          state.navigating = false
-          return
-        }
-      }
+    Log.error(`Route ${route.hash} not found`)
```
uguraslan: It seems we can't get `route.hash` if `route==false`.

src/router/router.js
```diff
@@ -640,15 +420,75 @@ const setOrAnimate = (node, transition, shouldAnimate = true) => {
...
+      state.navigating = false
+      return false
+    }
+  }
+}
+
+const executeAfterHook = async function (hooks, hookName, route, previousRoute) {
+  if (hooks && hooks[hookName]) {
+    try {
+      await hooks.afterEach.call(
```
uguraslan: is `hooks.afterEach` intentional here or should it be like `await hooks[hookName]` like `executeBeforeHook` does?

michielvandergeest: good one! fixed :)

src/router/router.js
michielvandergeest: should be fixed in: [#b8f2b5](https://github.com/lightning-js/blits/pull/650/changes/b8f2b5e806a660cf6d0f14aeafd70865643da08a)

src/router/router.js
```diff
@@ -250,316 +99,246 @@ const makeRouteObject = (route, overrides) => {
...
-          preventHashChangeNavigation = true
-          currentRoute = previousRoute
-          window.history.back()
-
-          navigatingBack = false
-          state.navigating = false
-          return
-        }
-      }
+    Log.error(`Route ${hash} not found`)
```
uguraslan: `getHash()` returns object, so maybe we need to use a key here `hash.hash`

src/router/utils.js
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Check if a value is an object
+ * @param {any} v
+ * @returns {boolean} True if v is an object
+ */
+export const isObject = (v) => typeof v === 'object' && v !== null
+
+/**
+ * Check if a value is a function
```
uguraslan: minor jsdoc issue.

src/router/router.js
```diff
@@ -250,316 +99,246 @@ const makeRouteObject = (route, overrides) => {
...
+
+  // apply out out transition on previous view if available, unless
+  // we're reusing the prvious page component
+  // FIX: truthy guard — previousRoute can be `false` (see history-push comment above).
+  if (previousRoute && reuse === false) {
+    // only animate when there is a previous route
+    shouldAnimate = true
+    let oldView = this[symbols.children].splice(1, 1).pop()
+    if (oldView) {
+      executeTransition(previousRoute.transition.out, oldView[symbols.holder], true)
```
uguraslan, line 296: I see `await executeTransition` in the rest of the code but not here, is this intentional?

src/router/router.js
```diff
@@ -250,316 +99,246 @@ const makeRouteObject = (route, overrides) => {
...
+        overrideOptions.keepAlive !== undefined
+          ? overrideOptions.keepAlive
+          : previousRoute.options && previousRoute.options.keepAlive
+
+      // cache the page when it's as 'keepAlive' instead of destroying
+      if (
+        navigatingBack === false &&
+        previousRoute.options &&
+        keepAlive === true &&
+        route.options.inHistory === true
```
uguraslan: should this be `previousRoute.options.inHistory` instead of `route.options.inHistory`?

src/router/router.js
michielvandergeest: yes, it's intentional

uguraslan (approved):
LGTM

PR #654, opened 2026-04-07
Marking Component As Non-Interactive If Hover/Unhover Hooks Are Not Exists

suresh-gangumalla:
Component's hoverable  configuration property will alter holder node `interactive` flag

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -23,7 +23,6 @@ import { Log } from '../../lib/log.js'
 import symbols from '../../lib/symbols.js'
 import Settings from '../../settings.js'
 import shaders from '../../lib/shaders/shaders.js'
-import { hasHook } from '../../lib/hooks.js'
 
 const holderComponentMap = new WeakMap()
 
@@ -529,8 +528,7 @@ const propsTransformer = {
     }
   },
   set holder(v) {
-    const identifier = this.element.component[symbols.identifier]
-    this.props['interactive'] = hasHook('hover', identifier) || hasHook('unhover', identifier)
+    this.props['interactive'] = v
   },
 }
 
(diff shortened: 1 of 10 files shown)


michielvandergeest:
hey @suresh-gangumalla, this looks good! But it does mean that we're introducing a somewhat arbitrary component config prop.

I had another idea, and I'm not sure how feasible it is, but it would make this whole thing pretty "automagic": what if we keep the generator the same (i.e. setting holder prop to true for all components) and then in the `holder` prop we verify if the component has a `hover` (or `unhover`, would that make sense to track separately?) hook specified. And if it does the component is considered to be `hoverable`

can you look into that approach?

src/engines/L3/element.js
```diff
@@ -534,7 +529,15 @@ const propsTransformer = {
     }
   },
   set holder(v) {
-    this.props['interactive'] = v
+    const identifier =
+      this.element && this.element.component
+        ? this.element.component[symbols.identifier]
+        : undefined
```
michielvandergeest: what are the use cases where `this.element` and `this.element.component` are `undefined?`

src/engines/L3/element.js
```diff
@@ -534,7 +529,15 @@ const propsTransformer = {
...
+    if (identifier !== undefined) {
+      this.props['interactive'] = hasHook('hover', identifier) || hasHook('unhover', identifier)
+    } else {
+      this.props['interactive'] = v
+    }
```
michielvandergeest: should we not skip this `else` block?

src/lib/hooks.js
```diff
@@ -57,3 +57,7 @@ export const registerHooks = (hooks = {}, identifier) => {
     if (typeof hooks[hook] === 'function') cbs[identifier][hook] = hooks[hook]
   })
 }
+
+export const hasHook = (hook, identifier) => {
+  return !!(cbs[identifier] && cbs[identifier][hook])
```
michielvandergeest: maybe rewrite to `cbs[identifier] !== undefined && hook in cbs[identifier]` ?

PR #657, opened 2026-04-08
create internal keyb event to work around browser issues

milesnash:
Some of the finer STBs that Blits is currently running on in production are running chrome 56, where there appears to be a bug with KeyboardEvent where the constructor doesn't apply many of the options provided in the 2nd argument, including `keyCode`, which Blits relies on internally.

This PR introduces a workaround class that extends KeyboardEvent and ensures the `keyCode` supplied in the constructor is applied.  Other props can be added as needed.


michielvandergeest (approved):
LGTM! ❤️

PR #663, opened 2026-04-15
Custom Plugin Documentation

suresh-gangumalla:
1. Explains different ways of creating custom plugins
2. Way to Registe custom plugins
3. Explains creating reactive state in custom plugins


michielvandergeest (approved):
LGTM!

PR #667, opened 2026-04-23
apply reactive :color to custom component holders

il-sairamg:
Fix reactive :color on custom components by generating holder updates the same way as other reactive attributes (removes the codegen skip that broke backgrounds.


michielvandergeest (approved):
LGTM

PR #671, opened 2026-05-08
refactor: move Sprite rendering to native L3 sprite texture path

il-sairamg:
Refactored sprite handling by removing the standalone Sprite component and moving sprite texture/frame resolution into the native L3 engine path.


michielvandergeest (commented):
LGTM

PR #673, opened 2026-05-12
fix: reset `preventHashChangeNavigation` on cancelled navigation and fix `reuseComponent` test

uguraslan:
Two fixes for the refactored router:

  1. `preventHashChangeNavigation` was only reset at the end of a completed navigation. When a before/beforeEach hook cancels navigation, the flag stayed `true` permanently, making all subsequent navigations to fail. Fixed by resetting the flag in the early return paths.
  2. Updated the `reuseComponent` test to seed a dummy view at index 1 to handle `currentRoute` state leaking from previous tests. (by @suresh-gangumalla )

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -140,7 +140,10 @@ export const navigate = async function () {
     previousRoute,
     currentPath
   )
-  if (beforeEachResult === false) return
+  if (beforeEachResult === false) {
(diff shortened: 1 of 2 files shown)


michielvandergeest (approved):
LGTM

PR #676, opened 2026-05-22
Fix reactivity guard computed parsing and  hoisting

il-sairamg:
Fix reactivity guard computed parsing and $appState hoisting


michielvandergeest (commented):
looks good, but I'm going to rely on @uguraslan to help me review this

also, it would be good if we can add test cases for the different use case, so we are sure what is supported and that it's not broken in future changes

michielvandergeest (approved):
LGTM

PR #688, opened 2026-06-17
Fixes the reactivity guard for long form function names

uguraslan:
`myValue: function() {...}` computed props never get a reactivity guard, unlike the shorthand `myValue() {...}` form. 

This change improves the existing check to match also with long form function definitions.


suresh-gangumalla (approved):
LGTM

PR #691, opened 2026-06-26
Feature/prevent cleanup attribute

wouterlucas:
Exposes the renderer's `textureOptions.preventCleanup` flag as a first-class `preventCleanup` attribute on Blits `<Element>` nodes.

When set to `true`, the underlying renderer texture is never evicted from GPU memory by the automatic cleanup cycle, regardless of whether the element is on-screen or under memory pressure. Useful for frequently reused textures (e.g. persistent backgrounds, shared UI chrome) that must always be instantly available without a reload delay.

## Usage
```
<Element src="assets/bg.png" preventCleanup="true" />
<Element src="assets/bg.png" fit="cover" preventCleanup="true" />
<Element :src="$imageUrl" :preventCleanup="$isPersistent" />
```

### DIFF
--- a/src/engines/L3/element.js
+++ b/src/engines/L3/element.js
@@ -318,10 +318,8 @@ const propsTransformer = {
     const resizeMode = {}
 
     if (v === 'cover' || v === 'contain') {
-      this.props['textureOptions'] = {
-        ...this.props.props['textureOptions'],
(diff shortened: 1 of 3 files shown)


michielvandergeest:
closing in favour of https://github.com/lightning-js/blits/pull/692

PR #698, opened 2026-07-01
Added test verifying frameTick and idle renderer listeners are unregi…

il-sairamg:
- Added test verifying frameTick, idle and active renderer listeners are unregistered on destroy.


michielvandergeest (approved):
LGTM

PR #699, opened 2026-07-02
Release 2.7.0

michielvandergeest:
- Added `keepAlive` option to `src`-object to prevent a texture from being cleaned up by the renderer
- Abstracted announcer (and speechSynthesis) to be platform provided (with speechSynthesis as the built-in default)
- Added dedicated findByData methods to testing harness to select specific nodes in the tree
- Improved test cases around frameTick cleanup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 2.7.0
+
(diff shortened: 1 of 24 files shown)


suresh-gangumalla (approved):
LGTM

PR #702, opened 2026-07-21
Fix/multi target effect cleanup

michielvandergeest:
Fixed issue with effects not being cleaned up if they would have more than 1 target


suresh-gangumalla (approved):
LGTM

PR #707, opened 2026-07-23
fix: default router view, fallback view and warning if two no name ro…

MarcinG98:
## Summary

This PR fixes several issues that occurred when using multiple `RouterView` instances, particularly around navigation state isolation and router view resolution.

### 1. Per-`RouterView` Navigation State (`src/router/router.js`)

**Problem**

Several navigation-related variables (`overrideOptions`, `navigationData`, `navigatingBack`, `navigatingBackTo`, `previousFocus`, and `preventHashChangeNavigation`) were stored as module-level singletons. Since all `RouterView` instances respond to hash changes, state from one view could inadvertently affect another, causing issues such as skipped `before` hooks or incorrect navigation behaviour.

**Solution**

* Replaced the module-level state with a `Map` (`perViewState`) keyed by router view name.
* Each `RouterView` now retrieves its own isolated state via `getViewState(this.name)`.
* State is automatically cleaned up when a `RouterView` is unregistered.

---

### 2. Default `RouterView` Fallback (`src/component/base/router.js`)

**Problem**

When multiple `RouterView` instances were registered and `router.to()` was called without specifying a view name, `getSingleRegisteredRouterView()` returned `null`. The existing fallback relied on tree traversal, which could resolve to the wrong router view, preventing the application's primary router hooks from executing.

**Solution**

* Added an intermediate lookup for the default (unnamed) `RouterView` (`name === ''`) before falling back to tree traversal.
* This ensures the application's primary router view is selected consistently.

---

### 3. Named `RouterView` Fallback (`src/component/base/router.js`)

**Problem**

Calling `this.$router.get('someName').to(...)` with a non-existent router view name resulted in navigation silently failing because resolution returned `null`.

**Solution**

* If a named router view cannot be found via the registry or tree traversal, navigation now falls back to the default (unnamed) `RouterView` instead of failing.

---

### 4. Warning for Multiple Unnamed `RouterView`s (`src/router/router.js`)

**Problem**

If multiple unnamed `RouterView` instances are registered, they share the same `perViewState` entry and ambiguous routing behaviour, effectively recreating the original issue.

**Solution**

* Added a `Log.warn()` during `registerRouterView()` when a second unnamed `RouterView` is registered.
* The warning advises developers to assign unique `name` props to each `RouterView` to avoid unexpected behaviour.

### DIFF
--- a/docs/router/basics.md
+++ b/docs/router/basics.md
@@ -91,6 +91,8 @@ In some cases you may want to have multiple `<RouterView />` components in your 
 
 By assigning a `name` to a `<RouterView />`, you can retrieve a Router instance scoped to that Router view using the `this.$router.get()`-method.
 
+Router view names must be unique and stable for the lifetime of each view. The unnamed RouterView uses the empty string as its name, so an application should have at most one unnamed RouterView. Duplicate or dynamically changed names can cause ambiguous view resolution and shared navigation state.
+
 ```js
 export default Blits.Application({
   template: `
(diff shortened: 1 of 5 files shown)


michielvandergeest (approved):
LGTM 🧇

PR #711, opened 2026-07-30
fix: destroy orphaned views when keepAlive is true but no route

AndresMonk:
If keepAlive flag is true but route is not found in the history, then we need to cleanup the nodes. Currently it is possible for the flag to be true but the route has been unrouted. This is potentially leaving orphaned nodes causing an increase in memory consumption the more navigations happen

### DIFF
--- a/src/router/router.js
+++ b/src/router/router.js
@@ -342,7 +342,9 @@ export const navigate = async function () {
           ? viewState.overrideOptions.keepAlive
           : this.previousRoute.options && this.previousRoute.options.keepAlive
 
-      // cache the page when it's as 'keepAlive' instead of destroying
+      // cache the page when it's marked as 'keepAlive' instead of destroying,
+      // but only when navigating forward AND the route is in history (so it can be restored on back)
+      let cached = false
       if (
         viewState.navigatingBack === false &&
         keepAlive === true &&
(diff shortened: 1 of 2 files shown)


src/router/router.js
```diff
@@ -354,15 +356,17 @@ export const navigate = async function () {
...
       /* Destroy the view in the following cases:
        * 1. Navigating forward, and the previous route is not configured with "keep alive" set to true.
-       * 2. Navigating back, and the previous route is configured with "keep alive" set to true.
-       * 3. Navigating back, and the previous route is not configured with "keep alive" set to true.
+       * 2. Navigating forward, keepAlive is true but inHistory is false (cannot be cached, would orphan).
+       * 3. Navigating back, and the previous route is configured with "keep alive" set to true.
+       * 4. Navigating back, and the previous route is not configured with "keep alive" set to true.
        */
-      if (this.previousRoute.options && (keepAlive !== true || viewState.navigatingBack === true)) {
+      if (!cached) {
```
michielvandergeest: this change drops the `viewState.navigatingBack` check entirely as it's not captured in the new `cached` variable - are we sure that's not needed anymore? could this be a potential regression?

michielvandergeest (changes requested):
thanks! looks good

I left a comment / question

also the router has quite some tests, would be great if we could add a test case proving it was previously broken (and now fixed with this change :) )

michielvandergeest (approved):

PR #726, opened 2026-08-17
Moved removeEffects function before redirecting focus to parent to pr…

michielvandergeest:
…event accidential $hasFocus effects to still run.


suresh-gangumalla (approved):
LGTM

PR #730, opened 2026-08-17
Aligned relying on eol property to block callbacks after destroy. Ali…

michielvandergeest:
…gned resetting node property to undefined instead of null.


suresh-gangumalla (approved):
LGTM

PR #731, opened 2026-08-17
Fix/reject fetchjson

michielvandergeest:
Improved error handling and parse rejection for fetchJson method


suresh-gangumalla (approved):
LGTM

PR #733, opened 2026-08-18
Fixed issue where a keyup callback could fire after component destroy…

michielvandergeest:
… and ensured keyup callbacks are cleaned up.


suresh-gangumalla (approved):
LGTM

PR #734, opened 2026-08-18
Fix application context default

il-sairamg:
apply computed type fallback to application context


michielvandergeest (commented):
LGTM

michielvandergeest (approved):

PR #762, opened 2026-09-18
added textCalculated listener to trigger layout changes

jfboeve:
- [ ] requires Renderer updated


src/engines/L3/element.js
```diff
@@ -680,6 +681,12 @@ const Element = {
 
     if (this.config.parent.props !== undefined && this.config.parent.props.__layout === true) {
       this.config.parent.triggerLayout(this.config.parent.props)
+      if (isTextNode === true) {
+        this.node.on('textCalculated', () => {
+          if (this.eol === true) return
+          this.config.parent.triggerLayout(this.config.parent.props)
+        })
+      }
```
michielvandergeest, line 694: doesn't this mean that for text nodes we'll trigger the layout twice in a row? First when text calculated, and then when loaded is fired. And in theory it should be the same size then, right?

michielvandergeest (approved):
LGTM

PR #763, opened 2026-09-18
made mouse move throttle configurable

johanneszelger:
mouse move throttle was a fixed 100ms, can now be configurate.

motivation: for some screens such as a login screen with a custom keyboard, 100ms can feel unresponsive.


src/application.js
```diff
@@ -149,7 +147,7 @@ const Application = (config) => {
 
     // limit the amount of move events per time frame
     mouseMoveHandler = (e) => {
-      if (e.timeStamp - lastMoved < 100) return
+      if (e.timeStamp - lastMoved < Settings.get('mouseMoveThrottle', 100)) return
```
michielvandergeest, line 152: can we move the `Settings.get()` method outside of this callback function, so we don't get the static session on every execution?

michielvandergeest (commented):
hey @johanneszelger, thanks for this PR! ❤️ 

Looks good, I left a small comment.

Also, we'll need the contributor licence agreement to be signed to move this one along. Thx

johanneszelger:
HI @michielvandergeest 

i did sign it, and i did add my github accounts email to the commits (that was the earlier force push). not sure why its not working?

johanneszelger:
ok signed again, i think it worked now

michielvandergeest:
nice! yeah CLA seems to be okay now 👍 

would be great if you could address this comment here: https://github.com/lightning-js/blits/pull/763/#discussion_r4046944384

johanneszelger:
Hi Michiel, 

i fixed the linting and added another improvement that allows consumers to change the trottling at runtime, for example during a fullscreen animation. Not sure if this is something you are happy to do, if not, i can revert the commit.

michielvandergeest:
> Hi Michiel,
> 
> i fixed the linting and added another improvement that allows consumers to change the trottling at runtime, for example during a fullscreen animation. Not sure if this is something you are happy to do, if not, i can revert the commit.

hey @johanneszelger yeah I understand what you're trying to do, and I understand that there are cases where you'd want to change the throttle ms runtime. The only implications of this is that we need to call `Settings.get('mouseMoveThrottle', 100))` upon every mouse move, while the value most of the times will stay static. That may impact performance, I fear

I think for now I prefer to hoist `Settings.get()` out of the `moveMouseHandler` and add runtime updating of settings (as a general construct) in a follow up PR - I have some ideas :)

michielvandergeest (approved):
LGTM

PR #764, opened 2026-09-18
fixed issue where switching between keyboard/mouse can lead to hovering not being recognized

johanneszelger:
## Problem

When i hover over something with my mouse, it gets focussed. When i then shift focus using my arrow keys, another item gets focussed. when i now wiggle my mouse over the first item, it won't get focus, even though i am back in mouse mode. 

## Cause

Blits remembers the currently hovered node to speed detection up, when the node is still hovered, there is no need  to check other nodes. However, when a key press clears the hover, Blits still remembers the node. So moving the pointer on that same node hits the "same node, nothing changed" pattern, even though the hover hook should fire again as the focus was lost.

## Fix
Reset current node when clearing Hover


michielvandergeest (approved):
hey @johanneszelger, nice catch! This is a use case we missed

We've verified the changes on our end and this is good to go

PR #766, opened 2026-09-21
Feat/allow rgba numeric colors

danielbrava:
I am adding functionality to do animations with colors to our blits animation plugin and wanted to make it more efficient. So I added option to allow numeric values for colors directly (like `0xffffff`)

Also made export of the html colors, so can be used for convenience in app or in plugin (prevent for duplicating)


michielvandergeest (approved):
LGTM

PR #768, opened 2026-09-21
Release 2.10.0

michielvandergeest:
- Fixed / improved type definitions
- Fixed issue with accidental sideeffect running when `$hasFocus` triggers after destroy
- Fixed issue where an error in an effect can break execution of other effects
- Added support for reactivity of `array.length`
- Improved reactivity in the case of nested array operations
- Fixed case where transition callbacks could run after destroy
- Fixed case where keyup callback could fire after component destroy
- Improved error handling and parse rejection for fetchJson method
- Fixed typing of `fpsUpdate`
- Added textCalculated listener to trigger layout changes
- Added `mouseMoveThrottle` setting to configure responsiveness of mouse movements
- Fixed issue where hover is not recognized after keyboard use
- Added support for numeric rgba colors
- Bumped renderer to 3.5.0


suresh-gangumalla:
LGTM

suresh-gangumalla (commented):
LGTM

PR #785, opened 2026-09-23
Feat/select method optimizations

danielbrava:
I found the $select method is not very optimal

It runs the full logic every time you call it, even with the same ref passed. Also it continues to itterate over the childeren when a ref is already found

In this PR I add early return when child with ref is found and add the element to cache for next time use

Tested with normal refs and refs in a forloop (cache is invalidated to prevent old matches returned)


src/component/base/methods.js
```diff
@@ -197,26 +194,59 @@ export default {
...
-            if (child[k]['ref'] === ref) selected = child[k]
-          })
+          const selected = Object.values(child).find((c) => c['ref'] === ref)
+          if (selected !== undefined) {
+            if (cache === undefined) {
+              cache = new Map()
+              selectCache.set(this, cache)
+            }
+            cache.set(ref, selected)
+            return selected
```
michielvandergeest, line 229: I'm seeing some duplication of code here. Can we consolidate this?

michielvandergeest (changes requested):
Nice one! Optimising the select method has been on my list for a while, but never got enough priority. So happy this got picked up 🙂 

There seems to be a merge conflict that needs to be fixed. And I've left a comment.

Thanks!

src/component/base/methods.js
danielbrava: Yes good point!

For readibilty it can be extract to a helper function, but then we have context change to a closure. I think extra function allocation maybe is less good for performance? That's why I did the small duplication

But I can change if you prefer

michielvandergeest: okay, yeah that makes sense - I think we can keep the intentional duplication for the sake of performance

let's keep as is 👍

src/lib/codegenerator/generator.js
```diff
@@ -532,6 +534,7 @@ const generateForLoopCode = function (templateObject, parent) {
     let to${forStartCounter}
 
     forloops[${forStartCounter}] = (collection = [], elms, created) => {
+      ${invalidateSelectCache ? "rootComponent?.[Symbol.for('invalidateSelectCache')]?.()" : ''}
```
michielvandergeest: just noticed this one

I prefer we don't use the optional chaining pattern (as it transpiles to very convoluted es5 code)

michielvandergeest (approved):
LGTM

